PR #13, opened 2018-02-09
Add Java Test Runner and Maven to the Pack

akaroml:


package.json
```diff
@@ -30,6 +30,8 @@
     },
     "extensionDependencies": [
         "redhat.java",
-        "vscjava.vscode-java-debug"
+        "vscjava.vscode-java-debug",
```
yaohaizh: How about readme?

akaroml: Next change list. I'll need the GIFs.

PR #34, opened 2018-08-17
Set the correct vscode engine compatibiltiy

sandy081:
vscode engine should be set to `^1.26` because older VS Code versions does not support `extensionPack` property

@akaroml Please have this change before publishing the extension


yaohaizh:
Thanks.

PR #53, opened 2018-10-31
Record toggleOverviewVisibility event

akaroml:


src/overview/index.ts
```diff
@@ -6,9 +6,17 @@ import * as path from 'path';
...
 
 const readFile = util.promisify(fsReadFile);
+
+import { instrumentOperation } from "vscode-extension-telemetry-wrapper";
+
 let overviewView: vscode.WebviewPanel | undefined;
 const KEY_SHOW_WHEN_USING_JAVA = 'showWhenUsingJava';
 
+const toggleOverviewVisibilityOperation = instrumentOperation("toggleOverviewVisibility", (operationId: string, context: vscode.ExtensionContext, visibility: boolean) => {
+  // TODO: record the actual visiblity value
```
akaroml: This requires the telemetry lib to provide sendInfo. @Eskibear

Eskibear: Discussed with @testforstephen , we can export such a general interface. It guards the `eventName` to be "**publisher.extension/info**", but the users should specify dimensions and measurements by their own. 
```ts
sendInfo(operationId: string, dimensions: { [key: string]: string }, measurements: { [key: string]: number }
```

PR #63, opened 2018-11-19
In-app recommendation: XML for pom.xml

akaroml:


src/recommendation/xml.ts
```diff
@@ -0,0 +1,31 @@
...
+
+import { validateAndRecommendExtension } from "../utils";
+import { instrumentCommand } from "../command";
+
+const KEY_XML_RECOMMENDATION_LAST_SHOW_TIME = "xmlRecommendationLastShowTime";
+
+export function initialize (context: vscode.ExtensionContext) {
+  const recommendationHandler = instrumentCommand(context, "recommendExtension", extensionRecommendationHandler);
+  context.subscriptions.push(vscode.workspace.onDidOpenTextDocument(e => {
+    if (e.uri.path.toLowerCase().indexOf("pom.xml") !== -1) {
```
testforstephen: => `if (e.fileName === "pom.xml")`

src/utils/index.ts
```diff
@@ -0,0 +1,16 @@
+import * as vscode from "vscode";
+
+export async function validateAndRecommendExtension(extName: string, message: string): Promise<boolean> {
+  const ext = vscode.extensions.getExtension(extName);
+  if (ext) {
+    return true;
+  }
+
+  const action = "Details";
```
testforstephen: Should add an `Install` button in the recommend dialog.

src/recommendation/xml.ts
Eskibear: `e.filename.toLowerCase()` for robustness? I've seen `POM.XML` in some projects.

src/utils/index.ts
akaroml: There is no API in vscode which allows you to "Install" an extension directly. So the workaround is to show the "Details" page and then you can click the "Install" button to achieve that.

src/recommendation/xml.ts
akaroml: e.fileName is the full path. === sign cannot be used directly to check whether the current file is pom.xml

src/utils/index.ts
testforstephen: VS Code's recommendation has the `Install` button, could we ask for their implementation?

PR #66, opened 2018-11-21
Refactor: consolidate all commands

akaroml:


src/extension.ts
```diff
@@ -1,28 +1,19 @@
...
-  context.subscriptions.push(vscode.commands.registerCommand('java.helper.createMavenProject', instrumentCommand(context, 'java.helper.createMavenProject', createMavenProjectCmdHanlder)));
-
-  context.subscriptions.push(vscode.commands.registerCommand('java.helper.createSpringBootProject', instrumentCommand(context, 'java.helper.createSpringBootProject', createSpringBootProjectCmdHandler)));
-
-  context.subscriptions.push(vscode.commands.registerCommand('java.helper.showExtension', instrumentCommand(context, 'java.helper.showExtension', showExtensionCmdHandler)));
-
-  context.subscriptions.push(vscode.commands.registerCommand('java.helper.openUrl', instrumentCommand(context, 'java.helper.openUrl', openUrlCmdHandler)));
-
   initUtils(context);
+  initCommands(context);
```
Eskibear: Just remind that previous the logic was before `initUtils`,  please make sure the order change won't block anything by your own.

akaroml: Nice comment. I'm considering removing the needs of initUtils because utilities are supposed to be pure functions. thanks.

PR #67, opened 2018-11-22
 Status Bar button to show overview page

akaroml:
image

Add this little entry to make the overview page entry more visible.


Eskibear:
What if you open a .java file?  Will there be two "Java" buttons?

VSCode has one to specify the language:
image

akaroml:
image

Eskibear:
For me, `Java $(thumb) $(info) Java` looks wired. And I don't think by adding an extra word "Java" you can allure users to click. I never click the "Java" button unless I need to explicitly tell VS Code what language the file is written.

PR #90, opened 2019-01-23
Update command to create spring boot projects

Eskibear:
The command `spring.initializr.createProject` serves as an entry to create a Spring Boot projects. It allows users to select between Maven and Gradle, and executes the procedure accordingly.

pack-spring-create

### DIFF
--- a/src/commands/handler.ts
+++ b/src/commands/handler.ts
@@ -22,7 +22,7 @@ export async function createSpringBootProjectCmdHandler(context: vscode.Extensio
     return;
   }
 
-  await vscode.commands.executeCommand('spring.initializr.maven-project');
+  await vscode.commands.executeCommand('spring.initializr.createProject');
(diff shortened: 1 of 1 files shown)


src/commands/handler.ts
```diff
@@ -22,7 +22,7 @@ export async function createSpringBootProjectCmdHandler(context: vscode.Extensio
     return;
   }
 
-  await vscode.commands.executeCommand('spring.initializr.maven-project');
+  await vscode.commands.executeCommand('spring.initializr.createProject');
```
yaohaizh: Need extract the command ID as constant values. Otherwise, it is easy to break the logic.

Eskibear: good suggestion, but I don't want to do it in this PR. Suggest to extract all other command names together in another separate PR.

PR #99, opened 2019-02-20
Update dependency versions

akaroml:


package.json
```diff
@@ -60,25 +60,25 @@
   },
   "devDependencies": {
     "@types/jquery": "^3.3.24",
-    "@types/node": "^8.10.38",
+    "@types/node": "^8.10.40",
```
yaohaizh: Should use node V10 since VSCode is depends on V10 already

Eskibear: Then do we need to also update the engine to `^1.31.0` ? Node/Electron/VSCode version maps can be seen [here](https://github.com/Sneezry/vscode-version-watcher)

testforstephen: This value specify the minimum compatiable vscode version. Don't suggest to update the `engine` if the extension doesn't use any new API only available in the new vscode.

Eskibear: I agree with your point. But my another thought is, by specifying engine e.g. `^1.20.0`, meaning we are declaring the extension works well with v1.20.0. But actually we haven't tested new features with VSCode legacy versions.

akaroml: The contract here is semantic version. It's VS Code's responsibility to not break backward-compatibility. From extension POV, my take is to require lower runtime version as possible to be able to work on a broader range of vscode distributions. The version upgrade needs to happen when we use some of the features that is only available in a newer version of vscode.

Eskibear: Yes, I totally agree what both of you have said.

akaroml: Thanks, please help sign off.

PR #114, opened 2019-03-15
Deserialize overview in new session

akaroml:


src/extension.ts
```diff
@@ -21,6 +21,8 @@ async function initializeExtension(operationId: string, context: vscode.Extensio
   initRecommendations(context);
   initMisc(context);
 
+  vscode.window.registerWebviewPanelSerializer('java.overview', new OverviewViewSerializer());
```
jdneo: Since `registerWebviewPanelSerializer()` will return a `Disposable`, we need to push it into the `context.subscriptions`

akaroml: Good suggestion!

akaroml:
Updated the PR to add the new disposable to subscriptions.

src/overview/index.ts
```diff
@@ -45,32 +46,42 @@ export async function overviewCmdHandler(context: vscode.ExtensionContext, opera
...
-  });
+function onDidDisposeWebviewPanel() {
+  overviewView = undefined;
+}
+
+async function initializeOverviewView(context: vscode.ExtensionContext, webviewPanel: vscode.WebviewPanel, onDisposeCallback: () => void) {
+  webviewPanel.iconPath = vscode.Uri.file(path.join(context.extensionPath, 'logo.lowres.png'));
+  webviewPanel.webview.html = await loadHtmlContent(context);
+
+  webviewPanel.onDidDispose(onDisposeCallback);
```
jdneo: Oh, sorry I missed this part. Since the Listener `onDidDispose()` is a `Disposable` as well, pushing into the subscription is needed.

You can take a reference here: https://github.com/jdneo/vscode-leetcode/blob/master/src/leetCodeResultProvider.ts#L22-L24

src/overview/index.ts
```diff
@@ -45,32 +46,42 @@ export async function overviewCmdHandler(context: vscode.ExtensionContext, opera
...
   });
 
-  overviewView.webview.postMessage({
+  webviewPanel.webview.postMessage({
     command: 'setOverviewVisibility',
     visibility: context.globalState.get(KEY_SHOW_WHEN_USING_JAVA)
   });
 
-  overviewView.webview.onDidReceiveMessage((e) => {
+  webviewPanel.webview.onDidReceiveMessage((e) => {
```
jdneo: Same as above

src/overview/index.ts
```diff
@@ -45,32 +46,42 @@ export async function overviewCmdHandler(context: vscode.ExtensionContext, opera
...
     if (e.command === 'setOverviewVisibility') {
       toggleOverviewVisibilityOperation(context, e.visibility);
     }
   });
 }
 
+async function loadHtmlContent(context: vscode.ExtensionContext) {
+  let buffer = await readFile(path.join(context.extensionPath, './out/assets/overview/index.html'));
```
jdneo: I prefer to use `context.asAbsolutePath()` here. Using `path.join()` is fine also. Not a bit deal.

akaroml: This does produce cleaner code! thanks.

src/overview/index.ts
akaroml: I actually found more. Thanks for the reminder!

PR #115, opened 2019-03-19
Recommend JDK when missing

akaroml:
Here's how the change looks.
image


src/extension.ts
```diff
@@ -9,6 +9,8 @@ import { initialize as initCommands } from "./commands";
 import { initialize as initRecommendations } from "./recommendation";
 import { initialize as initMisc, showReleaseNotesOnStart } from "./misc";
 import { showOverviewPageOnActivation, OverviewViewSerializer } from './overview';
+import { validateJavaRuntime } from "./java-runtime";
+import { valid } from 'semver';
```
jdneo: Is this member used anywhere?

.vscode/extensions.json
```diff
@@ -2,6 +2,5 @@
 	// See http://go.microsoft.com/fwlink/?LinkId=827846
 	// for the documentation about the extensions.json format
 	"recommendations": [
-		"eg2.tslint"
```
yaohaizh: use eslint?

src/java-runtime/index.ts
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+import findJavaHome = require("find-java-home");
+
+export async function validateJavaRuntime(): Promise<boolean> {
```
yaohaizh: This part of logic is already in language server. Should we avoid duplicate?

.vscode/extensions.json
Eskibear: `eg2.tslint`  has been deprecated in favor of the `ms-vscode.vscode-typescript-tslint-plugin`

src/java-runtime/index.ts
akaroml: The purpose of this change is to make the info very visible and intuitive.

src/extension.ts
akaroml: interesting...
The auto-completion put it here.

.vscode/extensions.json
akaroml: This is to remove the workspace recommendation.

src/java-runtime/index.ts
yaohaizh: For VSCode Java, it can directly use the java.home user setting. If user specified in the VSCode, it doesn't required additional JDK. The find home is based on JAVA_HOME value without knowledg of VSCode itself.

akaroml:
The pull request was updated to provide complete logic to search for JDK installations. The logic is the same to how Java LS' version.

akaroml:
> I would strongly suggest to change the toast's title to **_Java Development Kit Required_** because a _Java Runtime_ might actually already be present and it may cause confusion to use "Runtime" in the title and then talk about JDK in the body.

Thanks for the suggestion. The change is in the updated PR.

PR #119, opened 2019-03-29
Suggest JDK distribution by AdoptOpenJDK

akaroml:
This is to provide more JDK options for users to choose.
OS and architecture are automatically detected.

Screen Shot 2019-03-29 at 10 18 03 PM


src/overview/assets/index.ts
```diff
@@ -2,19 +2,38 @@
 // Licensed under the MIT license.
 
 import * as $ from "jquery";
-import "./index.scss";
+import './index.scss';
```
Eskibear: you are using **single** quote here

src/overview/assets/index.ts
```diff
@@ -2,19 +2,38 @@
...
+import "bootstrap/js/src/tab";
```
Eskibear: you are using **double** quote here

src/overview/assets/index.ts
```diff
@@ -2,19 +2,38 @@
...
 
 window.addEventListener("message", event => {
   if (event.data.command === "hideInstalledExtensions") {
     hideInstalledExtensions(event.data.installedExtensions);
     hideEmptySections();
-  } else if (event.data.command === "setOverviewVisibility") {
-    $("#showWhenUsingJava").prop("checked", event.data.visibility);
-  } else if (event.data.command === "showJavaRuntimePanel") {
-    $("#javaRuntimePanel").removeClass("d-none");
+  } else if (event.data.command === 'setOverviewVisibility') {
```
Eskibear: you are using **single** quote here

src/java-runtime/index.ts
```diff
@@ -4,8 +4,9 @@
 import * as vscode from "vscode";
 import * as cp from "child_process";
 import * as path from "path";
-import * as expandTilde from "expand-tilde";
+import expandTilde = require("expand-tilde");
```
Eskibear: is this change for tslint check?

src/java-runtime/index.ts
```diff
@@ -77,3 +78,21 @@ export async function validateJavaRuntime() {
...
+  let os: string = process.platform;
+  if (os === "win32") {
+    os = "windows";
+  } else if (os === "darwin") {
+    os = "mac";
+  } else {
+    os = "linux";
+  }
+
+  let arch = process.arch;
```
Eskibear: arch is unchanged

src/overview/assets/index.html
```diff
@@ -15,20 +15,97 @@ <h1>Java Overview</h1>
...
-          </ul>
-          <p>When you are finished, please reload Visual Studio Code to activate Java language server.</p>
-          <p class="mb-0">
-            <a class="btn btn-primary" href="command:workbench.action.reloadWindow" role="button" title="Reload Visual Studio Code">Reload Window</a>
-            <a class="btn btn-secondary" href="command:java.helper.openUrl?%22https%3A%2F%2Fgithub.com%2Fredhat-developer%2Fvscode-java%23setting-the-jdk%22" role="button" title="More info on how JDK is resolved">Having Trouble?</a>
-          </p>
+        <div class="card">
+          <div class="card-body">
+            <h4 class="alert-heading">⚠️ Java Development Kit Required</h4>
+            <p>Java Developerment Kit (JDK) 8 or later is required for developing Java applications. Please follow the links below to download and install JDK:
```
jdneo: typo: Developerment

src/overview/assets/index.html
```diff
@@ -15,20 +15,97 @@ <h1>Java Overview</h1>
...
+              </div>
+              <div class="tab-pane fade" id="otherJdkPanel" role="tabpanel" aria-labelledby="otherJdkTab">
+                <ul id="jdkProviderList" class="list-unstyled">
+                  <li><a href="command:java.helper.openUrl?%22https%3A%2F%2Fwww.azul.com%2Fdownloads%2Fazure-only%2Fzulu%2F%22" title="Recommended for Microsoft Azure Cloud and Azure Stack applications">Azul Zulu Enterprise for Azure</a></li>
+                  <li><a href="command:java.helper.openUrl?%22https%3A%2F%2Fwww.oracle.com%2Ftechnetwork%2Fjava%2Fjavase%2Fdownloads%2Findex.html%22" title="Java SE Downloads by Oracle">Java SE Downloads by Oracle</a></li>
+                  <li><a href="command:java.helper.openUrl?%22https%3A%2F%2Fopenjdk.java.net%2Finstall%2F%22" title="OpenJDK">OpenJDK</a></li>
+                </ul>
+              </div>
+            </div>
+            <p>When you are finished, please reload Visual Studio Code to activate Java language server.</p>
```
jdneo: `Java language server` or `Language Support for Java`?

I prefer the latter one since that's the official name of the extension.

src/java-runtime/index.ts
jdneo: Be careful of using `process.arch`: https://nodejs.org/api/process.html#process_process_arch

> The process.arch property returns a string identifying the operating system CPU architecture for which the Node.js binary was compiled.

So it sometimes will not return the actual system arch. For example, running 32-bit VS Code on an 64-bit OS. See:
Screen Shot 2019-03-30 at 11 24 33 AM

Well, it might be a big problem though. If we want to solve this problem, seems that we can leverage this node module: https://www.npmjs.com/package/arch

akaroml: Nice! Thank you!

akaroml:
> Workflow looks ok. But with some nits in TS files, single quotes are still used, looks it hasn't been guarded well.

Single quote now only appears in webpack.config.js, which is out of the scope of ts.

PR #144, opened 2019-04-26
Fix Java Version Parsing

ThadHouse:
The Java Version Parsing did not work for OpenJDK 11. Updated to do exactly the parsing the redhat java extension does.

Also updates package-lock.json, because my npm install did that and they should match.

Closes #143

Issue #142, opened 2019-04-26
Re-using new OpenJDK recommendation in separate vscode extension


NikolasKomonen:
Hi,

I recently saw this tweet: https://twitter.com/brunoborges/status/1121669654580064256
crediting @akaroml. This seems like a good feature to have in the XML language server, could you point to where this is implemented.

Also separately, a recommendation to allow some functionality to set/bring them to the `java.home`
preference in the case that they already have a JDK installed.

akaroml:
Thank you for the suggestion on the `java.home` setting. Our plan is to provide a consolidated view of all the Java related settings across different extensions.

Regarding the implementation, it has two parts:

1. The logic that runs in node.js environment to perform the actual detection and suggestion https://github.com/Microsoft/vscode-java-pack/tree/master/src/java-runtime
2. The logic that runs in the WebView to present the message and interact with users https://github.com/Microsoft/vscode-java-pack/tree/master/src/overview/assets

ntotten:
I have the same question about reusing the detection. I think it would be great if the Java detection and download UI were its own extension that many extensions could depend on. We have this same issue in the Salesforce Extensions.

akaroml:
> Salesforce

The question here is whether a dedicated extension just for acquiring Java runtime is worthy.

We try to avoid that in general, so I was looking for common ground. I just checked the Salesforce extensions and it turns out none of them depends on any Java extensions. So I failed to find one.

If we were to release such an extension, what kind of API would you be looking for? It would be very helpful to understand that. @NikolasKomonen @ntotten

NikolasKomonen:
For the most part, we would be looking for an API to give us the java path.

In terms of functionality, I think maybe if our extension requested the java path through your API then your extension would check the most likely locations for the jdk/jre and give the user a choice to select the one they want (there may be multiple) and return that choice it to our extension.

Also somewhere in that request, the option to include/ignore jre's since we specifically need a jdk.

Overall this would make it a bit more seamless for the user when initially setting up a java based extension. Right now the hurdles are them having to leave VSCode to download a jdk, then finding the path and setting it up correctly in the preferences, finally after that is all done they are then able to use the extension.

From the users point of view, they would open up and XML file in our case for the XML Language Server and they would be prompted to download a jdk. After downloading a window would pop up with the path(s) to the jdk and they would click one. Then the server should start working.

akaroml:
[...]

PR #145, opened 2019-04-27
Use Java language extension to get Java Home

ThadHouse:
This means the extension and its checking will always match what the language extension requires.

There is one downside with this, in that this does require the redhat extension to fully start up. This can delay loading of this extension, as that requires model syncing, but maybe I can work with that extension to fix that issue.

Either this or #144  will fix JDK parsing. Which one the owners would prefer is up to them, I will close the not merged one.

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,12 +95,14 @@
     "webpack-cli": "^3.3.1"
   },
   "extensionPack": [
-    "redhat.java",
     "vscjava.vscode-java-debug",
     "vscjava.vscode-java-test",
     "vscjava.vscode-maven",
     "vscjava.vscode-java-dependency",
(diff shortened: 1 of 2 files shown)


akaroml:
@ThadHouse Thanks for your contribution. I reviewed both #144 and #145. #144 was already merged.

My concern regarding #145 is that it adds vscode-java to the extension dependency list of vscode-java-pack. This would stop the pack extension to run when the language server extension fails to load. We want to avoid that because the pack has the role to help the users recover from those type of errors.

I agree that the logic of locating JDK should be shared in some way. #142 is about this too. But activating the Java language server just for this purpose is too heavy. We'll keep the discussion open and see what we can do.

Closing this PR. Thank you again.

PR #154, opened 2019-05-29
Dedicated Java Runtime View

akaroml:


src/java-runtime/index.ts
```diff
@@ -9,6 +9,51 @@ import * as pathExists from "path-exists";
...
+
+  javaRuntimeView = vscode.window.createWebviewPanel("java.runtime", "Java Runtime", {
+    viewColumn: vscode.ViewColumn.One,
+  }, {
+    enableScripts: true,
+    enableCommandUris: true,
+    retainContextWhenHidden: true
+  });
+
+  await initializeJavaRuntimeView(context, javaRuntimeView, onDidDisposeWebviewPanel);
```
jdneo: `javaRuntimeView` is defined at line 14, then why we still need to pass it as a method arg?

src/java-runtime/index.ts
```diff
@@ -9,6 +9,51 @@ import * as pathExists from "path-exists";
...
+}
+
+function onDidDisposeWebviewPanel() {
+  javaRuntimeView = undefined;
```
jdneo: Shall we call `javaRuntimeView.dispose()` before set it to `undefined`?

akaroml: This event handler is called when the dispose() is called. So there's no need to call it again.

src/java-runtime/index.ts
akaroml: javaRuntimeView is defined outside but it does not stop the initializeJavaRuntimeView being implemented in a functional fashion. We try to only work with parameters and local variables.

PR #172, opened 2019-08-20
Add command: java.runtime

akaroml:

### DIFF
--- a/src/java-runtime/assets/index.html
+++ b/src/java-runtime/assets/index.html
@@ -3,6 +3,7 @@
 
 <head>
   <meta charset="UTF-8">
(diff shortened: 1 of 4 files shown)


src/java-runtime/assets/index.html
```diff
@@ -0,0 +1,119 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+  <meta charset="UTF-8">
```
jdneo: Add CSP header

akaroml: Explicitly enabled 'unsafe-inline' to unblock this PR.

PR #173, opened 2019-08-21
Display Java runtime entries

akaroml:


src/java-runtime/assets/java.runtime.entries.tsx
```diff
@@ -0,0 +1,40 @@
+import * as React from "react";
```
jdneo: license

src/java-runtime/index.ts
```diff
@@ -10,6 +10,10 @@ import * as request from "request-promise-native";
 import findJavaHome = require("find-java-home");
 import architecture = require("arch");
 import { loadTextFromFile, getExtensionContext } from "../utils";
+import { JavaRuntimeEntry, JavaRuntimeEntryTypes } from "./types";
+import _ = require("lodash");
```
jdneo: `import * as _ from 'lodash';`

src/java-runtime/types.ts
```diff
@@ -0,0 +1,14 @@
+export interface JavaRuntimeEntry {
```
jdneo: license

PR #175, opened 2019-08-22
Remove java runtime from overview

akaroml:

### DIFF
--- a/src/java-runtime/index.ts
+++ b/src/java-runtime/index.ts
@@ -183,8 +183,8 @@ export async function findJavaRuntimeEntries(): Promise<JavaRuntimeEntry[]> {
       entry.isValid = false;
       entry.hint = "This path is not pointing to a JDK.";
       const pathObject = path.parse(entry.path||"");
-      if (pathObject.name && pathObject.name === "bin") {
-        entry.hint += " \"bin\" should be removed from the path.";
+      if (pathObject.name && pathObject.name.toLowerCase() === "bin") {
+        entry.hint += " Try remove the \"bin\" from the path.";
       }
       return entry;
     }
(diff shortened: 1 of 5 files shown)


src/java-runtime/index.ts
```diff
@@ -175,14 +181,18 @@ export async function findJavaRuntimeEntries(): Promise<JavaRuntimeEntry[]> {
   return Promise.all(_.map(entries, async entry => {
     if (!await checkJdkInstallation(entry.path)) {
       entry.isValid = false;
-      entry.hint = "This path does not point to a JDK.";
+      entry.hint = "This path is not pointing to a JDK.";
+      const pathObject = path.parse(entry.path||"");
+      if (pathObject.name && pathObject.name === "bin") {
```
jdneo: nit: `pathObject.name === "bin"` is enough. Leave the choice to you

src/java-runtime/index.ts
```diff
@@ -175,14 +181,18 @@ export async function findJavaRuntimeEntries(): Promise<JavaRuntimeEntry[]> {
   return Promise.all(_.map(entries, async entry => {
     if (!await checkJdkInstallation(entry.path)) {
       entry.isValid = false;
-      entry.hint = "This path does not point to a JDK.";
+      entry.hint = "This path is not pointing to a JDK.";
+      const pathObject = path.parse(entry.path||"");
+      if (pathObject.name && pathObject.name.toLowerCase() === "bin") {
```
akaroml: Case insensitive comparison is what I really wanted in the first place...

src/java-runtime/index.ts
akaroml: Addressed. Please check the new commit.

src/java-runtime/index.ts
Eskibear: any concern of using `path.basename(string)` instead of `path.parse(string)`?

akaroml: works too. But let's move on...

PR #183, opened 2019-08-30
Getting Started - Quick Start

akaroml:


src/java-runtime/assets/index.ts
```diff
@@ -58,6 +58,7 @@ declare function acquireVsCodeApi(): any;
 const vscode = acquireVsCodeApi();
 
 function requestJdkInfo(jdkVersion: string, jvmImpl: string) {
+  console.log(`request ${jdkVersion} ${jvmImpl}`);
```
Eskibear: is the console.log on purpose?

akaroml: omg, seems like the merge had some problem

PR #189, opened 2019-09-01
Present first view according to user setting

akaroml:

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,25 @@
         "command": "java.gettingStarted",
         "title": "Java: Getting Started"
       }
-    ]
+    ],
+    "configuration":{
+      "title": "Java Pack Configuration",
+      "properties": {
+        "java.help.firstView": {
+          "type": [
+            "string"
+          ],
+          "enum": [
+            "auto",
+            "overview",
+            "gettingStarted"
+          ],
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -61,7 +61,25 @@
...
+            "string"
+          ],
+          "enum": [
+            "auto",
+            "overview",
+            "gettingStarted"
+          ],
+          "default": "auto",
+          "description": "Sets the default view which is presented during the first use.",
+          "scope": "window"
```
jdneo: Any thinking about setting the scope to `window` here?

akaroml: Windows (instance) specific settings which can be configured in user, workspace, or remote settings.

By doing so, users are able to choose the level of impact of this setting, and it sounds appropriate to me. Other thoughts?

jdneo: To me, it seems more like a `application` scope setting. When will the user needs to set different first view for different workspaces?

akaroml: There could be more first UX views targeting a different audience. E.g. we want to present a view for Spring Boot developers. We provide a sample workspace and set the default view to Spring Getting Started, and this option allows us to do that at the workspace level.

jdneo: I see

PR #258, opened 2019-11-20
Add SapMachine as JDK option

RealCLanger:
With this proposed change, SapMachine - SAP's OpenJDK build - will be added to the JDK options in VS Code.


akaroml:
Thanks!

RealCLanger:
Cool, thanks for merging.

Is there some vscode preview version where I can see the changed page?

Thanks
Christoph

PR #333, opened 2020-03-18
Add SonarLint to recommendation list

akaroml:


jdneo:
@henryju this is the PR to recommend sonarlint in Java Pack extension

PR #457, opened 2020-07-28
Extension Guide - Sync Installation Status

akaroml:

An option is checked and disabled when the corresponding extension is already installed.

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3683,7 +3683,7 @@
         },
         "readable-stream": {
           "version": "2.3.6",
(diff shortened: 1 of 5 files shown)


akaroml:
> I have two questions:
> 
> * **Project Manager for Java**  is installed but not detected. Not sure if it's because of the display name?
> * It cannot distinguish "Disabled" and "Not Installed". And what's worse, I cannot install a disabled extension by clicking "Install Selected(1)". Nothing happens.
> 
> image

This PR is about showing the installation status. The actual installation will be done in future PRs.

PR #462, opened 2020-07-31
Show Extension Guide in Experiment Group

akaroml:


src/exp/index.ts
```diff
@@ -0,0 +1,37 @@
+import * as vscode from "vscode";
```
jdneo, line 1: nit: license

src/utils/extension.ts
```diff
@@ -0,0 +1,12 @@
+export function getExtensionName() {
+  const packageInfo = getPackageInfo();
```
jdneo, line 2: nit: license

PR #473, opened 2020-08-20
Settings to reduce popups

### LINKED ISSUE #469: Preference setting not to show any **Help First View** page
First activation of the Java Extension Pack leads to Java Overview page opened.
We'd like not to open it on the first startup at all. However there is no such option... It's either `gettingStarted`, `auto` or `overview`. Can we have `none`?
I could provide a PR but seems like it is an easy thing to fix anyway. (Let me know if you'd prefer a PR for this)

BoykoAlex:
Fixes #469 

Settings not to show at startup:
 - overview/getting started page
 - release notes

'none' enum value is added for `firstView`... maybe instead of it another boolean on/off first view would be better - let me know please

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,15 +78,23 @@
           "enum": [
             "auto",
             "overview",
-            "gettingStarted"
+            "gettingStarted",
+            "none"
           ],
           "enumDescriptions": [
             "Automatically pick the first experience view",
             "Present the Java Overview page",
-            "Present the Java Getting Started page"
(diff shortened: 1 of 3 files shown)


Eskibear (approved):

Eskibear:
FYI that, in the future it's possible that we remove `java.help.showReleaseNotes`, and re-use vscode's setting `update.showReleaseNotes` instead. 
(currently it's not possible because it's of `application` scope)

PR #482, opened 2020-09-22
Add survey prompt logic

xqzlgy2:
This PR added survey prompt logic for Java extension pack.

First, 10% users were randomly picked for showing survey. Then, user's usage was divided into 3 stages according to their first activation time (0 to 7, 7 to 21, 21 days or later), and different questionnaire will be shown in different stage.

If user finished the survey, it won't be prompt in current stage (or 12weeks in final stage). If user choose "don't show again", the survey won't prompt. If user choose "remind me later", the survey will be disabled for 2.5 to 3.5 days randomly.

### DIFF
--- a/src/survey/constants.ts
+++ b/src/survey/constants.ts
@@ -2,34 +2,34 @@
 // Licensed under the MIT license.
 
 export const BannerActions = {
-    bannerLabelYes: 'Take Survey',
-    bannerLabelNo: 'Don\'t show again',
-    remindLater: 'Remind Me later'
+  bannerLabelYes: 'Take Survey',
+  bannerLabelNo: 'Don\'t show again',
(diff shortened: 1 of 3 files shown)


src/survey/constants.ts
```diff
@@ -0,0 +1,35 @@
...
+}
+
+export const SurveyBanner = {
+  message: 'Do you mind taking a quick feedback survey about the Java extension pack?',
+  action: [
+    BannerActions.bannerLabelYes,
+    BannerActions.bannerLabelNo,
+    BannerActions.remindLater
+  ],
+  FIRST_STAGE_URL: 'https://www.microsoft.com',
```
jdneo, line 17: @akaroml @yaojinyang Do we have the survey links for the three different stages?

PR #493, opened 2020-11-02
New view of configure java runtime

Eskibear:
Now it's split into 3 tabviews + 1 fixed panel below.

Tabviews:
* Project JDKs: configure JDK for projects, modifying mapping of source level to JDKs. 
* Setup Lanaguage Server: configure JDK to launch LS itself.
* Installed JDKs: an overview listing all installed JDKs and versions.

Fixed panel:
* Install A JDK: shortcut to download a new JDK.


Eskibear:
Screenshots for 3 tab views.

image

image

image

Eskibear:
TSLint failed because of `overview` page, probably introduced in previous changes. Gonna fix that first.

jdneo:
Some feedbacks after playing with the bit

1. I find here we have too large empty space
image
image


2. If user updates the language level, the configuration page will not auto refresh. Shall we listen to the [onDidClasspathUpdate event](https://github.com/redhat-developer/vscode-java/blob/master/src/extension.api.ts#L91)?

Eskibear:
1. I'll adjust and remove the spaces. 

2. I can give it a try. However a corner case is, when webview is open, if vscode-java extension is not yet activated, the event is invalid.

testforstephen:
This is the result i got when i opened spring-petclinic project.

image

It lists both 11 and 1.8, since my project just used Java 1.8 and the description says it's Java versions detected in my projects? This is a little confusing to me, is this a bug?

Eskibear:
@testforstephen Good catch. It's for the default project, which I didn't exclude from the list.

Eskibear:
@jdneo @testforstephen 
Comments addressed, can you take a look/give it a try?

Eskibear:
I've just added Workspace Overview, to provide a straightforward view of relationship between projects and Java versions. The layout looks like below. 
image

Further action items:
- [ ] Project name is fake (simply using base name of the project root folder), as we lack an API to get it. See https://github.com/eclipse/eclipse.jdt.ls/issues/1590. 
- [ ] Provide shortcut to jump to pom.xml/build.gradle if user wants to change Java version.

testforstephen:
> Project name is fake (simply using base name of the project root folder), as we lack an API to get it.

Yeah, finally the LS should provide an API to list the project metadata. So far, you could parse name from .project file, or leverage `java.project.list` command exposed by Project Explorer.  https://github.com/microsoft/vscode-java-dependency/blob/master/src/commands.ts

Eskibear:
Changelog:
* fetch project metadata from Project Manager, to show correct project names
* click `Maven`/`Gradle` to open pom.xml / build.gradle
* refine wording
 
Now it looks like:
image

jdneo:
This page can also be registered into the navigation bar of Java Projects explorer

jdneo (approved):
LGTM after playing with the page

PR #502, opened 2020-11-18
Fix: /usr is wrongly recognized as JDK in macOS Big Sur

Eskibear:

### DIFF
--- a/src/java-runtime/utils/findJavaRuntime.ts
+++ b/src/java-runtime/utils/findJavaRuntime.ts
@@ -91,7 +91,10 @@ async function fromPathEnv(): Promise<string[]> {
                 try {
                     buffer = cp.execSync(macUtility, { cwd: dir });
                     const absoluteJavaHome = "" + buffer.toString().replace(/\n$/, "");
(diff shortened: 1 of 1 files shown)


jdneo (approved):

src/java-runtime/utils/findJavaRuntime.ts
```diff
@@ -91,15 +91,24 @@ async function fromPathEnv(): Promise<string[]> {
...
                 }
             }
         }
     }
 
-    return ret;
+    if (isMac) {
+        // Exclude /usr, because in macOS Big Sur /usr/bin/javac is no longer symlink.
+        // See https://github.com/redhat-developer/vscode-java/issues/1700#issuecomment-729478810
+        return ret.filter(item => item !== "/usr");
```
testforstephen, line 108: item !== "/usr" && item != "/usr/" ?

Eskibear: `path.dirname` always trim the trailing "/", so I think it's not necessary. What do you think?

testforstephen (approved):

PR #503, opened 2020-11-19
align jdk validation with vscode-java

Eskibear:


Eskibear:
vscode-java has changed jdk detection approach. So here we align the logic accordingly, poping up the view only when vscode-java fails to find a valid jdk.

testforstephen (approved):

PR #508, opened 2020-11-20
fix: current LS JDK not selected on start

Eskibear:
javaHome was shaded.

### DIFF
--- a/src/java-runtime/utils/upstreamApi.ts
+++ b/src/java-runtime/utils/upstreamApi.ts
@@ -49,7 +49,11 @@ export async function resolveRequirements(): Promise<any> {
         }
 
         if (javaVersion < REQUIRED_JDK_VERSION) {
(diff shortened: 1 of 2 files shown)


src/java-runtime/utils/upstreamApi.ts
```diff
@@ -49,7 +49,11 @@ export async function resolveRequirements(): Promise<any> {
         }
 
         if (javaVersion < REQUIRED_JDK_VERSION) {
-            invalidJavaHome(reject, `Java ${REQUIRED_JDK_VERSION} or more recent is required to run the Java extension. (Current version: ${javaVersion}, JDK: ${javaHome})`);
+            let message = `Java ${REQUIRED_JDK_VERSION} or more recent is required to run the Java extension.`;
```
Eskibear, line 52: previously if `java.home` is not specified, and only JDK 8 is installed, then javaHome=null, javaVersion=0. The error message was confusing as "version: 0, JDK: null"

jdneo (approved):

PR #516, opened 2020-12-01
add settings for showing release notes after an update

Eskibear:
This PR adds a new setting `java.update.showReleaseNotes` to allow users to control whether to pop up release notes after an update. 

VS Code's setting `update.showReleaseNotes` is of `application` scope, see https://github.com/microsoft/vscode/blob/f74e473238aca7b79c08be761d99a0232838ca4c/src/vs/platform/update/common/update.config.contribution.ts#L47-L52

`application` scope makes more sense to me, but then it can only be configured in user settings. I don't know whether user settings can be preset in dev containers. Here I use `window` to ensure it can at least be overrided by workspace settings.

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,6 +88,12 @@
           "default": "auto",
           "description": "Sets the default view which is presented during the first use.",
           "scope": "window"
+        },
+        "java.update.showReleaseNotes": {
+          "type": "boolean",
+          "default": true,
+          "description": "Show Release Notes after an update.",
+          "scope": "window"
         }
       }
     }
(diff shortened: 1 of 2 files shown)


Eskibear:
FYI, `settings` section in file `devcontainer.json` can specify `machine` scope settings.  https://code.visualstudio.com/docs/remote/devcontainerjson-reference

Eskibear:
I'm closing it for the moment.   The ultimate way of doing this, is to reuse vscode's `update.showReleaseNotes` setting.

PR #525, opened 2020-12-14
fix #521: add null check before accessing javaExt.exports

### LINKED ISSUE #521: "Cannot read property 'onDidClasspathUpdate' of undefined" error when opening Java runtime guide
When opening the new Java runtime guide, the operation will throw error such as `Cannot read property 'onDidClasspathUpdate' of undefined`.

Eskibear:
fix #521

### DIFF
--- a/src/java-runtime/index.ts
+++ b/src/java-runtime/index.ts
@@ -139,7 +139,7 @@ async function initializeJavaRuntimeView(context: vscode.ExtensionContext, webvi
 
   // refresh webview with latest source levels when classpath (project info) changes
   const javaExt = vscode.extensions.getExtension("redhat.java");
-  if (javaExt && javaExt.isActive && javaExt.exports.onDidClasspathUpdate) {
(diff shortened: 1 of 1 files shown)


src/java-runtime/index.ts
```diff
@@ -139,7 +139,7 @@ async function initializeJavaRuntimeView(context: vscode.ExtensionContext, webvi
 
   // refresh webview with latest source levels when classpath (project info) changes
   const javaExt = vscode.extensions.getExtension("redhat.java");
-  if (javaExt && javaExt.isActive && javaExt.exports.onDidClasspathUpdate) {
+  if (javaExt && javaExt.isActive && javaExt.exports && javaExt.exports.onDidClasspathUpdate) {
```
jdneo, line 142: javaExt?.isActive && javaExt?.exports?.onDidClasspathUpdate

jdneo (approved):

Issue #540, opened 2021-02-08
New UX of welcome view


Eskibear:
Todo Items:

- [x] Implement a welcome page + tour page. In welcome page, provide shortcuts in group. In tour page, take a tour of main features. (PR #535)
- [x] only pop up tour page when open in first time. (PR #544)
- [x] switch to welcome page automatically after finishing the tour. (PR #541 )
- [x] assign proper screenshots in the tour page. (PR #546)

Eskibear:
Telemetry

- [ ] telemetry for actions (external links, button, tab-switching...)

When there's no open folder:
- [ ] Proper guide in Java Project Explorer, instead of an error message. @jdneo
- [ ] Proper guide/infor in Testing view. @jdneo

visibility of welcome view (@Eskibear )
- [ ] Hide in Codespaces by default.
- [ ] Controlled by A/B testing

PR #541, opened 2021-02-08
navigate to welcome page after finishing feature tour

Eskibear:
Part of #540 , in case users don't click the button, we navigate to welcome page after 5 seconds.

### DIFF
--- a/src/welcome/assets/components/TourPage.tsx
+++ b/src/welcome/assets/components/TourPage.tsx
@@ -17,35 +17,48 @@ export default class TourPage extends Component<{
         this.state = {
             step: 0
         };
+
+        this.steps = [
+            this.getStartingPage(),
+            ...this.getContentPages(),
+            this.getEndingPage()
+        ];
+
+        this.timer = undefined;
     }
 
+    private steps: JSX.Element[];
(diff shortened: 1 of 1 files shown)


jdneo (approved):

src/welcome/assets/components/TourPage.tsx
```diff
@@ -17,35 +17,48 @@ export default class TourPage extends Component<{
...
+            ...this.getContentPages(),
+            this.getEndingPage()
+        ];
+
+        this.timer = undefined;
     }
 
+    private steps: JSX.Element[];
+    // auto navigate to welcome page in last step
+    private timer: number | undefined;
```
jdneo, line 32: Shouldn't it declared as `NodeJS.Timer`?

Eskibear: `NodeJS.Timer` failed webpack compilation.  In Javascript, `settimeout` returns id of the timer, thus it's a `number`. In NodeJS, it's a `Timer`.  Here this is used in a webview, so `number` type in plain JS should be used.

jdneo: I see. That makes sense.

PR #557, opened 2021-03-05
fix webview command to execute

Eskibear:
Previously, **array of args** was wrongly passed as **the first arg** in `webviewCmdLinkHandler`.

### DIFF
--- a/src/utils/index.ts
+++ b/src/utils/index.ts
@@ -58,5 +58,10 @@ export async function webviewCmdLinkHandler(obj: { webview: string, identifier: 
     webview,
     identifier
   });
-  await vscode.commands.executeCommand(command, args);
+
+  if (args !== undefined) {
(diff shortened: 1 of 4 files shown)


src/utils/index.ts
```diff
@@ -58,5 +58,5 @@ export async function webviewCmdLinkHandler(obj: { webview: string, identifier:
     webview,
     identifier
   });
-  await vscode.commands.executeCommand(command, args);
+  args ? await vscode.commands.executeCommand(command, ...args) : await vscode.commands.executeCommand(command);
```
jdneo, line 61: `await vscode.commands.executeCommand(command, args ? ...args : undefined)`

Eskibear: done

jdneo (approved):

PR #567, opened 2021-03-15
Initialize the classpath configuration page

jdneo:
image

### DIFF
--- a/src/classpath/types.ts
+++ b/src/classpath/types.ts
@@ -1,0 +1,33 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+export interface ProjectInfo {
+    name: string;
+    rootPath: string;
+}
+
+export interface ClasspathComponent {
+    projectType: ProjectType;
+    sourcePaths: string[];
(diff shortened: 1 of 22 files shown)


src/classpath/types.ts
```diff
@@ -0,0 +1,33 @@
...
+}
+
+export interface ClasspathComponent {
+    projectType: ProjectType;
+    sourcePaths: string[];
+    defaultOutputPath: string;
+    referenceLibraries: string[];
+}
+
+export enum ProjectType {
```
Eskibear, line 16: same enum used in Runtime webview. Consider to extract...

src/classpath/types.ts
```diff
@@ -0,0 +1,33 @@
...
+    Default = "Default project",
+    UnmanagedFolder = "Unmanaged folder",
+    Maven = "Maven",
+    Gradle = "Gradle",
+    Others = "Others",
+}
+
+export enum NatureId {
```
Eskibear, line 24: same here

jdneo: Sure I'll extract them in the next PR.

Eskibear (approved):

PR #570, opened 2021-03-17
Extract getProjectType() utility

jdneo:

### DIFF
--- a/src/utils/jdt.ts
+++ b/src/utils/jdt.ts
@@ -1,0 +1,29 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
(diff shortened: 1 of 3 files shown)


src/utils/jdt.ts
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+import * as fse from "fs-extra";
+import * as path from "path";
+import { ProjectType } from "./webview";
```
Eskibear, line 6: `utils/jdt.ts` requires `ProjectType` from `utils/webview.ts`. It looks strange. maybe ProjectType should also belong to jdt.ts?

Eskibear (approved):

PR #577, opened 2021-03-19
Tune the UI styles for classpath configuration page

jdneo:
image


Eskibear (approved):

Eskibear:
By using chevronDownIcon, you should test whether it works well in light theme

jdneo:
Yes it is. Because the color is set to `color: var(--vscode-settings-textInputForeground);`

PR #580, opened 2021-03-23
Adopt to new API and check the dependency version

jdneo:
This PR have two part of changes.

1. Adopt the new API key (`defaultOutputPath` -> `outputPath`).
2. Check the version of `redhat.java`, and report error when the version is stale.

### DIFF
--- a/src/classpath/classpathConfigurationView.ts
+++ b/src/classpath/classpathConfigurationView.ts
@@ -19,7 +19,7 @@ let currentProjectRoot: vscode.Uri;
 const SOURCE_PATH_KEY: string = "org.eclipse.jdt.ls.core.sourcePaths";
 const OUTPUT_PATH_KEY: string = "org.eclipse.jdt.ls.core.outputPath";
 const REFERENCED_LIBRARIES_KEY: string = "org.eclipse.jdt.ls.core.referencedLibraries";
-const LEAST_JAVA_EXTENSION_VERSION: string = "0.77.0";
+const MINIMUM_JAVA_EXTENSION_VERSION: string = "0.77.0";
 
 export async function showClasspathConfigurationPage(context: vscode.ExtensionContext): Promise<void> {
     if (classpathConfigurationPanel) {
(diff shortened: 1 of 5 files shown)


src/classpath/classpathConfigurationView.ts
```diff
@@ -11,13 +11,15 @@ import minimatch from "minimatch";
...
+import compareVersions from "compare-versions";
 
 let classpathConfigurationPanel: vscode.WebviewPanel | undefined;
 let lsApi: LanguageServerAPI | undefined;
 let currentProjectRoot: vscode.Uri;
 const SOURCE_PATH_KEY: string = "org.eclipse.jdt.ls.core.sourcePaths";
-const OUTPUT_PATH_KEY: string = "org.eclipse.jdt.ls.core.defaultOutputPath";
+const OUTPUT_PATH_KEY: string = "org.eclipse.jdt.ls.core.outputPath";
 const REFERENCED_LIBRARIES_KEY: string = "org.eclipse.jdt.ls.core.referencedLibraries";
+const LEAST_JLS_VERSION: string = "0.77.0";
```
Eskibear, line 22: JLS is misleading. It's version of java extension instead of language server

src/classpath/classpathConfigurationView.ts
```diff
@@ -109,16 +111,27 @@ async function initializeWebview(context: vscode.ExtensionContext): Promise<void
...
         sendError(err);
         return false;
     }
+
+    if (compareVersions(javaExt.packageJSON.version, LEAST_JLS_VERSION) < 0) {
+        classpathConfigurationPanel?.webview.postMessage({
+            command: "onException",
+            exception: ClasspathViewException.StaleJavaExtension,
+        });
+        const err: Error = new Error("The extension version of 'redhat.java' is too stale.");
```
Eskibear, line 129: do you think we should also report current version if java extension?

jdneo: Yes I think so

Eskibear (dismissed):

Eskibear (approved):

Issue #604, opened 2021-04-21
Provide better format configuration experience


jdneo:
For Formatter Setting custom editor:
- [x] webview layout #607 
- [x] render whitespace #624 
- [x] enhancement based on current settings #626 
- [x] functionality of editor #628

PR #609, opened 2021-04-28
functionality for formatter settings page

CsCherrYY:
This is Part 2 of #604 .
Change of setting will trigger formatting in the mean time, also support undo and redo in the editor.

### DIFF
--- a/src/commands/index.ts
+++ b/src/commands/index.ts
@@ -31,12 +31,12 @@ export function initialize(context: vscode.ExtensionContext) {
   context.subscriptions.push(vscode.commands.registerCommand("java.gettingStarted", instrumentCommand(context, "java.gettingStarted", javaGettingStartedCmdHandler)));
   context.subscriptions.push(vscode.commands.registerCommand("java.extGuide", instrumentCommand(context, "java.extGuide", javaExtGuideCmdHandler)));
   context.subscriptions.push(instrumentOperationAsVsCodeCommand("java.webview.runCommand", webviewCmdLinkHandler));
-  context.subscriptions.push(vscode.commands.registerCommand("java.welcome", (options) => showWelcomeWebview(context, options)));
+  context.subscriptions.push(vscode.commands.registerCommand("java.welcome", instrumentCommand(context, "java.welcome", showWelcomeWebview)));
   const javaFormatterSettingsEditorProvider: JavaFormatterSettingsEditorProvider = new JavaFormatterSettingsEditorProvider(context);
   const editorOptions = { webviewOptions: {enableFindWidget: true, retainContextWhenHidden: true}, supportsMultipleEditorsPerDocument: false };
   context.subscriptions.push(vscode.window.registerCustomEditorProvider(JavaFormatterSettingsEditorProvider.viewType, javaFormatterSettingsEditorProvider, editorOptions));
-  context.subscriptions.push(vscode.commands.registerCommand("java.formatterSettings", () => javaFormatterSettingsEditorProvider.showFormatterSettingsEditor()));
-  context.subscriptions.push(vscode.commands.registerCommand("java.classpathConfiguration", async () => {
+  context.subscriptions.push(vscode.commands.registerCommand("java.formatterSettings", instrumentCommand(context, "java.formatterSettings", () => javaFormatterSettingsEditorProvider.showFormatterSettingsEditor())));
+  context.subscriptions.push(vscode.commands.registerCommand("java.classpathConfiguration", instrumentCommand(context, "java.classpathConfiguration", async () => {
     const showCustomizedView: boolean = await getExpService()?.getTreatmentVariableAsync(TreatmentVariables.VSCodeConfig, TreatmentVariables.CustomizedClasspathConfigurationView, true /*checkCache*/) || false;
     if (showCustomizedView) {
       showClasspathConfigurationPage(context);
@@ -46,5 +46,5 @@ export function initialize(context: vscode.ExtensionContext) {
       await vscode.commands.executeCommand("workbench.action.openWorkspaceSettings");
       markdownPreviewProvider.show(context.asAbsolutePath(path.join("webview-resources", "classpathConfiguration.md")), "Classpath Settings", context);
     }
-  }));
+  })));
 }
--- a/src/welcome/index.ts
+++ b/src/welcome/index.ts
@@ -10,7 +10,7 @@ const KEY_SHOW_WHEN_USING_JAVA = "showWhenUsingJava";
 const KEY_IS_WELCOME_PAGE_VIEWED = "isWelcomePageViewed";
 let welcomeView: vscode.WebviewPanel | undefined;
 
-export async function showWelcomeWebview(context: vscode.ExtensionContext, options?: any) {
+export async function showWelcomeWebview(context: vscode.ExtensionContext, _operationId?: string, options?: any) {
     if (options?.firstTimeRun) {
         setFirstTimeRun(context, true);
     }


src/formatter-settings/index.ts
```diff
@@ -4,27 +4,347 @@
...
+            webviewPanel.webview.postMessage({
+                command: "loadSetting",
+                setting: setting,
+            });
+        }
+    }
+
+    private async findJavaSpecificSetting<T>(setting: string): Promise<T | undefined> {
+        let result: T | undefined;
+        const config = vscode.workspace.getConfiguration().get<Object>("[java]");
```
jdneo, line 242: Why not use `{ languageId: "java" }` as scope

src/formatter-settings/index.ts
```diff
@@ -4,27 +4,347 @@
...
+        const documentDOM = new DOMParser().parseFromString(document.getText());
+        const profilesVersion = documentDOM.documentElement.getAttribute("version");
+        if (profilesVersion) {
+            this.settingsVersion = profilesVersion;
+        }
+        const profiles = documentDOM.documentElement.getElementsByTagName("profile");
+        if (!profiles || profiles.length === 0) {
+            throw new Error("No available profiles exist");
+        }
+        for (let i = 0; i < profiles.length; i++) {
```
jdneo, line 173: Why we use index-loop here?

src/formatter-settings/index.ts
```diff
@@ -4,27 +4,347 @@
...
+            }
+            const cloneElement = this.profileSettingsArray[this.profileSettingsArray.length - 1].cloneNode();
+            const originString: string = new XMLSerializer().serializeToString(cloneElement);
+            cloneElement.setAttribute("id", id);
+            cloneElement.setAttribute("value", value);
+            const newString: string = new XMLSerializer().serializeToString(cloneElement);
+            const edit: vscode.WorkspaceEdit = new vscode.WorkspaceEdit();
+            edit.insert(document.uri, new vscode.Position(cloneElement.lineNumber - 1, cloneElement.columnNumber - 1 + originString.length), EOL + " ".repeat(cloneElement.columnNumber - 1) + newString);
+            vscode.workspace.applyEdit(edit);
+            cloneElement.lineNumber++;
```
jdneo, line 278: Is it possible to not host the linNumber?

src/formatter-settings/index.ts
```diff
@@ -4,27 +4,347 @@
...
+    public async showFormatterSettingsEditor() {
+        if (!this.settingsUrl) {
+            vscode.window.showInformationMessage("No available Java Formatter Profile found, do you want to create a default profile?",
+                "Yes", "No").then((messageResult) => {
+                    if (messageResult === "Yes") {
+                        this.addFormatter();
+                    }
+                });
+        } else {
+            if (this.isRemote(this.settingsUrl)) {
```
jdneo, line 303: If thi's a remote URL, it's better to render it with read-only mode

src/formatter-settings/index.ts
```diff
@@ -4,27 +4,347 @@
...
+
+    private isRemote(path: string): boolean {
+        return path !== null && path.startsWith("http:/") || path.startsWith("https:/");
+    }
+
+    private async addFormatter(): Promise<void> {
+        const defaultProfile: string = path.join(this.context.extensionPath, "webview-resources", "formatter.xml");
+        let profilePath: string;
+        let relativePath: string | undefined;
+        const hasWorkspace: boolean = vscode.workspace.workspaceFolders !== undefined && vscode.workspace.workspaceFolders.length > 0;
```
jdneo, line 320: const workspaceFolder = vscode.workspace.workspaceFolders

And check the length of `workspaceFolder`

src/formatter-settings/index.ts
```diff
@@ -4,27 +4,347 @@
...
+        if (hasWorkspace) {
+            profilePath = path.join(vscode.workspace.workspaceFolders![0].uri.fsPath, ".vscode", "formatter.xml");
+            relativePath = path.join(".vscode", "formatter.xml");
+        } else {
+            const folder: string = path.join(this.context.extensionPath, "..", "vscjava.vscode-java-pack");
```
jdneo, line 325: Do not generate files into the extension installation path. If the version is updated, all these files will be missed

src/formatter-settings/index.ts
```diff
@@ -4,27 +4,347 @@
...
+                        this.addFormatter();
+                    }
+                });
+        } else {
+            if (this.isRemote(this.settingsUrl)) {
+                await vscode.window.showInformationMessage("The active formatter profile is in remote URL, please change to a local file and try to reopen this setting again.");
+                return;
+            }
+            const profilePath = await this.getProfilePath(this.settingsUrl);
+            vscode.commands.executeCommand("vscode.openWith", vscode.Uri.file(profilePath), "java.formatterSettingsEditor");
```
jdneo, line 308: What will happen if the profilePath does not exist?

src/formatter-settings/index.ts
```diff
@@ -4,27 +4,347 @@
...
+            command: "formattedContent",
+            content: content,
+        });
+    }
+
+    private async changeProfileSetting(id: string, value: string, document: vscode.TextDocument): Promise<void> {
+        const profileElement = this.profileElements.get(id);
+        if (!profileElement) {
+            // add a new setting not exist in the profile
+            if (!this.profileSettingsArray.length) {
```
jdneo, line 267: What will happen if the inserted setting in not supported in the current version of the profile

src/formatter-settings/assets/features/formatterSettings/FormatterSettingView.tsx
```diff
@@ -14,29 +14,9 @@ const FormatterSettingsView = (): JSX.Element => {
...
-      case String(Category.Whitespace):
-        dispatch(changeActiveCategory(Category.Whitespace));
-        break;
-      case String(Category.Wrapping):
-        dispatch(changeActiveCategory(Category.Wrapping));
-        break;
-      default:
-        break;
-    }
+    dispatch(changeActiveCategory(+element));
```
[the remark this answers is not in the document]
CsCherrYY: The value of element is the value of category as string type. For example, `Category.Common` is `0` but the element here will be `"0"`. So simply add `+` here to change to the origin enum value.

src/formatter-settings/index.ts
CsCherrYY: Here we just put the files into the path named `vscjava.vscode-java-pack` without version information, so the files will not be lost when the extension updates.

src/formatter-settings/assets/features/formatterSettings/FormatterSettingView.tsx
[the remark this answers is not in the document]
jdneo: It's better to use `Number("0")`

src/formatter-settings/index.ts
jdneo: Even it will not be lost, this path should not be used to persist any caches. Consider to use something like global storage

src/formatter-settings/index.ts
CsCherrYY: Since profiles are [HTMLCollection](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCollection), which doesn't support for-of loop, we have to change it to array by `Array.(profiles)` to support for-of loop, I'm afraid of the translating time. So I use index-loop here.

src/formatter-settings/assets/features/formatterSettings/FormatterSettingView.tsx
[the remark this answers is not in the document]
CsCherrYY: ok

src/formatter-settings/index.ts
CsCherrYY: Will use `path.join(this.context.globalStorageUri.fsPath)` as the target folder.

src/formatter-settings/index.ts
CsCherrYY: It will throw an unexpected error now. Will check if the profile exists.

src/formatter-settings/index.ts
CsCherrYY: Here we can use
```
const config = vscode.workspace.getConfiguration("editor", { languageId: "java" });
const tabSize = config.get("tabSize")
```
to get the value.

src/formatter-settings/index.ts
CsCherrYY: Will add a check for the inserted setting, if not supported in current version, will throw an error message.

src/formatter-settings/index.ts
CsCherrYY: I'm not sure if we should
1. Create a temporary xml file for the remote profile to open the custom editor
2. Keep it in read-only mode and remain access to the preview
Instead, if we want to support the remote URL, I suggest offering an option to guide (or help) the user to download the profile and change the setting. Does this make sense?

src/formatter-settings/index.ts
```diff
@@ -4,27 +4,347 @@
...
 import * as path from "path";
 import { loadTextFromFile } from "../utils";
+import * as fse from "fs-extra";
+import { FormatterConverter } from "./FormatterConverter";
+import { Example, initializeSupportedProfileSettings, initializeSupportedVSCodeSettings, JavaConstants, SupportedSettings } from "./FormatterConstants";
+import { DOMParser, XMLSerializer } from "xmldom";
+import { ExampleKind } from "./types";
+import compareVersions from "compare-versions";
+import { EOL } from "os";
 export class JavaFormatterSettingsEditorProvider implements vscode.CustomTextEditorProvider {
```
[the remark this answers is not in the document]
CsCherrYY: Will add keys for each setting.

src/formatter-settings/index.ts
jdneo: If the setting is set to a remote URL by user on purpose, that's indicating that the user doesn't want to change the content.

So, read-only mode is required. Meanwhile, an option to let user be able to download the file and customize it would be a plus!

src/formatter-settings/index.ts
```diff
@@ -4,27 +4,347 @@
...
 
-    public async showFormatterSettingsEditor() {
-        // Use a fake profile
-        const defaultProfile: string = path.join(this.context.extensionPath, "webview-resources", "formatter.xml");
-        const resource = vscode.Uri.file(defaultProfile);
-        vscode.commands.executeCommand("vscode.openWith", resource, "java.formatterSettingsEditor");
-    }
-
-    public async resolveCustomTextEditor(_document: vscode.TextDocument, webviewPanel: vscode.WebviewPanel, _token: vscode.CancellationToken): Promise<void> {
+    public async resolveCustomTextEditor(document: vscode.TextDocument, webviewPanel: vscode.WebviewPanel, _token: vscode.CancellationToken): Promise<void> {
```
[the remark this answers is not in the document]
CsCherrYY: Will add an event/message from the frontend to tell the backend for initializing.

src/formatter-settings/index.ts
```diff
@@ -4,27 +4,347 @@
...
     constructor(private readonly context: vscode.ExtensionContext) {
+        this.profileElements = new Map<string, any>();
+        this.profileSettings = new Map<string, string>();
+        this.profileSettingsArray = [];
+        this.settingsVersion = JavaConstants.JAVA_FORMATTER_SETTINGS_VERSION;
+        this.settingsUrl = vscode.workspace.getConfiguration("java").get<string>(JavaConstants.SETTINGS_URL_KEY);
+        this.settingsProfile = vscode.workspace.getConfiguration("java").get<string>(JavaConstants.SETTINGS_PROFILE_KEY);
+        context.subscriptions.push(vscode.workspace.onDidChangeConfiguration(e => {
+            if (e.affectsConfiguration(`java.${JavaConstants.SETTINGS_URL_KEY}`)) {
+                this.settingsUrl = vscode.workspace.getConfiguration("java").get<string>(JavaConstants.SETTINGS_URL_KEY);
```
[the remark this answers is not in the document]
CsCherrYY: Will send a notification to let the user choose if the webview should be reloaded.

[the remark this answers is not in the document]
jdneo: How much time it takes for paring the format xml?

src/formatter-settings/index.ts
```diff
@@ -237,19 +256,10 @@ export class JavaFormatterSettingsEditorProvider implements vscode.CustomTextEdi
...
-        for (const entry of entries) {
-            if (entry[0] === setting) {
-                result = entry[1];
-            }
-        }
-        return result;
+    private async getSetting<T>(setting: string): Promise<T | undefined> {
+        const config = vscode.workspace.getConfiguration(undefined, { languageId: "java" });
+        const result = config.get<T>(setting);
+        return result ? result : vscode.workspace.getConfiguration().get<T>(setting);
```
jdneo, line 287: return config.get<T>(setting) || vscode.workspace.getConfiguration().get<T>(setting)

src/formatter-settings/index.ts
```diff
@@ -216,13 +235,13 @@ export class JavaFormatterSettingsEditorProvider implements vscode.CustomTextEdi
...
-        const javaTabSize: number | undefined = await this.findJavaSpecificSetting<number>("editor.tabSize");
-        const javaSpacePolicy: boolean | undefined = await this.findJavaSpecificSetting<boolean>("editor.insertSpaces");
-        const vscodeTabSize: number | undefined = vscode.workspace.getConfiguration().get<number>("editor.tabSize");
-        const vscodeSpacePolicy: boolean | undefined = vscode.workspace.getConfiguration().get<boolean>("editor.insertSpaces");
-        const tabSize = (javaTabSize === undefined) ? vscodeTabSize : javaTabSize;
-        const spacePolicy = (javaSpacePolicy === undefined) ? vscodeSpacePolicy : javaSpacePolicy;
-        for (const setting of initializeSupportedVSCodeSettings()) {
+        const tabSize: number | undefined = await this.getSetting<number>("editor.tabSize");
+        const spacePolicy: boolean | undefined = await this.getSetting<boolean>("editor.insertSpaces");
+        if (tabSize === undefined || spacePolicy === undefined) {
```
jdneo, line 265: Give them a default value instead of undefined. Meanwhile, log the message when undefined happens

PR #626, opened 2021-05-14
Define messages for formatter setting editor

CsCherrYY:
this PR includes the following things about formatter setting editor:
- Introduce some fixes for current code base.
- Define messages between backend and frontend including related changes. A fake provider in the backend to support this for the temporary use.


src/commands/index.ts
```diff
@@ -33,7 +33,7 @@ export function initialize(context: vscode.ExtensionContext) {
   context.subscriptions.push(instrumentOperationAsVsCodeCommand("java.webview.runCommand", webviewCmdLinkHandler));
   context.subscriptions.push(vscode.commands.registerCommand("java.welcome", instrumentCommand(context, "java.welcome", showWelcomeWebview)));
   const javaFormatterSettingsEditorProvider: JavaFormatterSettingsEditorProvider = new JavaFormatterSettingsEditorProvider(context);
-  const editorOptions = { webviewOptions: {enableFindWidget: true, retainContextWhenHidden: true}, supportsMultipleEditorsPerDocument: false };
+  const editorOptions = { webviewOptions: {enableFindWidget: true, retainContextWhenHidden: true}};
```
CsCherrYY, line 36: currently `supportsMultipleEditorsPerDocument` is only available for `CustomReadonlyEditorProvider` and `CustomEditorProvider`, so this option is removed.

src/formatter-settings/assets/style.scss
```diff
@@ -34,6 +34,12 @@ $enable-rounded: false;
   }
 }
 
+@media (max-width: 992px) and (min-width: 783px) {
```
CsCherrYY, line 37: Here to fix the bug when a setting section has an odd number of settings, in the layout where a line contains two settings, the last setting would be larger than others.

src/formatter-settings/assets/features/formatterSettings/FormatterSettingView.tsx
```diff
@@ -77,8 +73,8 @@ const FormatterSettingsView = (): JSX.Element => {
       <Row className="flex-grow-1 d-flex flex-nowrap view-body">
         <Col className="flex-grow-0">{naviBar}</Col>
         <Col className="d-flex view-content">
-          <Row className="flex-nowrap flex-column flex-lg-row d-flex w-100 h-100">
-            <Col className="flex-grow-0 setting-container d-flex flex-row flex-lg-column flex-lg-nowrap">{<Setting/>}</Col>
+          <Row className="flex-grow-1 flex-nowrap flex-column flex-lg-row d-flex w-100 h-100">
```
CsCherrYY, line 76: to make the distance between the preview and the right border the same as that between the category and the left border.

src/formatter-settings/index.ts
```diff
@@ -26,5 +31,65 @@ export class JavaFormatterSettingsEditorProvider implements vscode.CustomTextEdi
...
+                }
+                default:
+                    break;
+            }
+        });
+    }
+
+    private async formatWithProfileSettings(webviewPanel: vscode.WebviewPanel) {
+        // fake provider: use currently default settings
+        const content = await vscode.commands.executeCommand<string>("java.execute.workspaceCommand", "java.edit.stringFormatting", Example.getExample(this.exampleKind), JSON.stringify([]), JavaConstants.CURRENT_FORMATTER_SETTINGS_VERSION);
```
CsCherrYY, line 87: See: https://github.com/eclipse/eclipse.jdt.ls/pull/1702 for the API

src/commands/index.ts
jdneo: This file is used to register command, please consider move the provider registration to other place

src/formatter-settings/FormatterConstants.ts
```diff
@@ -5,10 +5,17 @@ import { Category, ExampleKind, JavaFormatterSetting, ValueKind } from "./types"
 
 export namespace JavaConstants {
     export const JAVA_CORE_FORMATTER_ID = "org.eclipse.jdt.core.formatter";
+    export const CURRENT_FORMATTER_SETTINGS_VERSION = "21";
```
jdneo, line 8: Do we have better way to handle the version? If a new format version comes, you need to release a new version to handle with that.

src/formatter-settings/assets/features/formatterSettings/components/Setting.tsx
```diff
@@ -28,28 +32,37 @@ const Setting = (): JSX.Element => {
...
 
-  const generateSetting = (setting: JavaFormatterSetting) => {
+  const handleClick = (exampleKind: ExampleKind) => {
+    dispatch(activateExampleKind({ exampleKind: exampleKind }));
+  };
+
+  const generateSetting = (setting: JavaFormatterSetting): JSX.Element => {
     if (!setting.name || !setting.id || !setting.value) {
-      return;
+      return (<></>);
```
jdneo, line 48: Can we just return null here?

src/formatter-settings/assets/features/formatterSettings/components/Setting.tsx
```diff
@@ -28,28 +32,37 @@ const Setting = (): JSX.Element => {
...
+          <div className="setting-section" key={`${setting.id}`} onClick={() => handleClick(setting.exampleKind)}>
             <Form.Check type="checkbox" className="pl-0" id={`${setting.id}`} >
               <Form.Check.Input type="checkbox" checked={setting.value === "true"} onChange={handleChangeCheckbox} />
               <Form.Check.Label className="setting-section-description">
                 <Icon className="codicon" icon={checkIcon} />
                 <div className="setting-section-description-checkbox">{setting.name}.</div>
               </Form.Check.Label>
+              <br></br>
+              <span className="warning">
+                {willBeOverriden ? "When detecting, the indentation settings will be overriden based on file contents." : ""}
```
jdneo, line 64: When enabled, ....

src/formatter-settings/assets/features/formatterSettings/formatterSettingViewSlice.ts
```diff
@@ -39,26 +40,39 @@ export const formatterSettingsViewSlice = createSlice({
...
+      state.vscodeSettings = action.payload.setting;
+      for (const setting of state.vscodeSettings) {
+        if (setting.id === SupportedSettings.TABULATION_SIZE) {
+          document.documentElement.style.setProperty("--vscode-tab-size", setting.value);
+        } else if (setting.id === VSCodeSettings.DETECT_INDENTATION) {
+          state.detectIndentation = (setting.value === "true");
         }
       }
     },
+    activateExampleKind: (state, action) => {
```
jdneo, line 58: This reducer is wired. Why sending a request to the server in a reducer?

src/formatter-settings/index.ts
```diff
@@ -26,5 +31,65 @@ export class JavaFormatterSettingsEditorProvider implements vscode.CustomTextEdi
         };
         const resourceUri = this.context.asAbsolutePath("./out/assets/formatter-settings/index.html");
         webviewPanel.webview.html = await loadTextFromFile(resourceUri);
+        this.exampleKind = ExampleKind.INDENTATION_EXAMPLE;
+        webviewPanel.webview.onDidReceiveMessage(async (e) => {
+            switch (e.command) {
+                case "onWillInitialize": {
```
jdneo, line 37: No need to wrap each case use `{}`

src/formatter-settings/index.ts
```diff
@@ -26,5 +31,65 @@ export class JavaFormatterSettingsEditorProvider implements vscode.CustomTextEdi
...
+                    this.formatWithProfileSettings(webviewPanel);
+                    break;
+                }
+                case "onWillChangeExampleKind": {
+                    this.exampleKind = e.exampleKind;
+                    this.formatWithProfileSettings(webviewPanel);
+                    break;
+                }
+                case "onWillChangeSetting": {
+                    // fake provider: replace the setting directly
```
jdneo, line 56: what's the meaning of fake provider?

src/formatter-settings/assets/features/formatterSettings/components/Setting.tsx
CsCherrYY: Yeah, I find that `null` is a better practice from [Preventing Component from Rendering](https://reactjs.org/docs/conditional-rendering.html#preventing-component-from-rendering).

src/formatter-settings/index.ts
CsCherrYY: Here we just change the setting in memory instead of modifying the file content. I'll change the description to make it clear.

src/commands/index.ts
CsCherrYY: Can we move the provider to `extension.ts` as an exported variable? Since we need `context` to initialize the provider and use `provider.showFormatterSettingsEditor()` to register the command.

src/formatter-settings/FormatterConstants.ts
CsCherrYY: Here we can mark it as an engineering debt to find a proper solution later.

src/formatter-settings/assets/features/formatterSettings/formatterSettingViewSlice.ts
CsCherrYY: Will remove requests in the reducer.

jdneo:
CI failed

jdneo (approved):

PR #628, opened 2021-05-17
feat: Add ability to modify the formatter profile

CsCherrYY:
This PR includes the following functionalities:
- Read settings from formatter profile and show them in the editor.
- Changes on the editor will take effect on the active profile.
- Push diagnostics to show the found errors in the current profile.
- Support Read-Only Mode in remote profile

### DIFF
--- a/src/formatter-settings/assets/features/formatterSettings/components/Setting.tsx
+++ b/src/formatter-settings/assets/features/formatterSettings/components/Setting.tsx
@@ -30,6 +30,7 @@ const Setting = (): JSX.Element => {
     if (!value || value === "") {
       value = "0";
     }
+    value = Number(value);
     onWillChangeSetting(id, value);
   };
 
--- a/src/formatter-settings/assets/features/formatterSettings/components/Setting.tsx
+++ b/src/formatter-settings/assets/features/formatterSettings/components/Setting.tsx
@@ -37,7 +37,7 @@ const Setting = (): JSX.Element => {
   };
 
   const generateSetting = (setting: JavaFormatterSetting) => {
-    if (!setting.name || !setting.id || !setting.value) {
+    if (!setting.name || !setting.id) {
       return null;
     }
     const candidates = [];
--- a/src/formatter-settings/assets/features/formatterSettings/components/Setting.tsx
+++ b/src/formatter-settings/assets/features/formatterSettings/components/Setting.tsx
@@ -25,9 +25,7 @@ const Setting = (): JSX.Element => {
   };
 
   const handleChangeInput = (e: any) => {
-    const id = e.target.id;
-    let value = e.target.value;
-    onWillChangeSetting(id, value);
+    onWillChangeSetting(e.target.id, e.target.value);
   };
 
   const handleSelect = (setting: JavaFormatterSetting, entry: string) => {
--- a/src/formatter-settings/assets/features/formatterSettings/components/Setting.tsx
+++ b/src/formatter-settings/assets/features/formatterSettings/components/Setting.tsx
@@ -27,10 +27,6 @@ const Setting = (): JSX.Element => {
   const handleChangeInput = (e: any) => {
     const id = e.target.id;
     let value = e.target.value;
-    if (!value || value === "") {
-      value = "0";
-    }
-    value = Number(value);
     onWillChangeSetting(id, value);
   };
 
--- a/src/formatter-settings/assets/features/formatterSettings/components/Setting.tsx
+++ b/src/formatter-settings/assets/features/formatterSettings/components/Setting.tsx
@@ -17,6 +17,7 @@ const Setting = (): JSX.Element => {
   const vscodeSettings: JavaFormatterSetting[] = useSelector((state: any) => state.formatterSettings.vscodeSettings);
   const activeCategory: Category = useSelector((state: any) => state.formatterSettings.activeCategory);
   const detectIndentation: boolean = useSelector((state: any) => state.formatterSettings.detectIndentation);
+  const readOnly: boolean = useSelector((state: any) => state.formatterSettings.readOnly);
 
   const handleChangeCheckbox = (e: any) => {
     const id = e.target.id;
@@ -47,7 +48,7 @@ const Setting = (): JSX.Element => {
         return (
           <div className="setting-section" key={`${setting.id}`} onClick={() => handleClick(setting.exampleKind)}>
             <Form.Check type="checkbox" className="pl-0" id={`${setting.id}`} >
-              <Form.Check.Input type="checkbox" checked={setting.value === "true"} onChange={handleChangeCheckbox} />
+              <Form.Check.Input type="checkbox" checked={setting.value === "true"} onChange={handleChangeCheckbox} disabled={readOnly} />
               <Form.Check.Label className="setting-section-description">
                 <Icon className="codicon" icon={checkIcon} />
                 <div className="setting-section-description-checkbox">{setting.name}.</div>
@@ -74,7 +75,7 @@ const Setting = (): JSX.Element => {
           <div className="setting-section" key={`${setting.id}`} onClick={() => handleClick(setting.exampleKind)}>
             <span className="setting-section-description">{setting.name}.</span>
             <Dropdown className="mt-1">
-              <Dropdown.Toggle className="dropdown-button flex-vertical-center text-left">
+              <Dropdown.Toggle className="dropdown-button flex-vertical-center text-left" disabled={readOnly}>
                 <span>{setting.value}</span>
                 <Icon className="codicon" icon={chevronDownIcon} />
               </Dropdown.Toggle>
@@ -88,7 +89,7 @@ const Setting = (): JSX.Element => {
         return (
           <div className="setting-section" key={`${setting.id}`} onClick={() => handleClick(setting.exampleKind)}>
             <Form.Label className="setting-section-description my-0">{setting.name}.</Form.Label>
-            <Form.Control className="pl-1 mt-1" type="number" id={setting.id} value={setting.value} onChange={handleChangeInput}></Form.Control>
+            <Form.Control className="pl-1 mt-1" type="number" id={setting.id} value={setting.value} onChange={handleChangeInput} disabled={readOnly}></Form.Control>
           </div>
         );
       default:
--- a/src/formatter-settings/index.ts
+++ b/src/formatter-settings/index.ts
@@ -12,7 +12,7 @@ import { Example, getSupportedVSCodeSettings, JavaConstants, SupportedSettings, 
 import { FormatterConverter } from "./FormatterConverter";
 import { RemoteProfileProvider } from "./RemoteProfileProvider";
 import { DOMElement, ExampleKind, ProfileContent } from "./types";
-import { addDefaultProfile, downloadFile, getProfilePath, getTargetPath, getVersion, getVSCodeSetting, isRemote, openFormatterSettings, parseProfile } from "./utils";
+import { addDefaultProfile, downloadFile, getProfilePath, getAbsoluteTargetPath, getVSCodeSetting, isRemote, openFormatterSettings, parseProfile } from "./utils";
 export class JavaFormatterSettingsEditorProvider implements vscode.CustomTextEditorProvider {
 
     public static readonly viewType = "java.formatterSettingsEditor";
@@ -54,7 +54,7 @@ export class JavaFormatterSettingsEditorProvider implements vscode.CustomTextEdi
     }
 
     public async showFormatterSettingsEditor(): Promise<void> {
-        if (!await this.checkProfileSettings() || !this.settingsUrl) {
+        if (this.webviewPanel || !await this.checkProfileSettings() || !this.settingsUrl) {
             return;
         }
         const filePath = this.readOnly ? vscode.Uri.parse(this.settingsUrl).with({ scheme: RemoteProfileProvider.scheme }) : vscode.Uri.file(this.profilePath);
@@ -72,7 +72,7 @@ export class JavaFormatterSettingsEditorProvider implements vscode.CustomTextEdi
             this.webviewPanel = webviewPanel;
         }
 
-        webviewPanel.webview.options = {
+        this.webviewPanel.webview.options = {
             enableScripts: true,
             enableCommandUris: true,
         };
@@ -125,7 +125,7 @@ export class JavaFormatterSettingsEditorProvider implements vscode.CustomTextEdi
                             });
                         return;
                     }
-                    webviewPanel.dispose();
+                    this.webviewPanel?.dispose();
                     await this.downloadAndUse(settingsUrl);
                     break;
                 }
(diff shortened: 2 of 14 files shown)


src/formatter-settings/utils.ts
```diff
@@ -0,0 +1,61 @@
...
+import * as fse from "fs-extra";
+import * as path from "path";
+import * as vscode from "vscode";
+import { instrumentOperation, sendInfo } from "vscode-extension-telemetry-wrapper";
+
+export async function getProfilePath(formatterUrl: string): Promise<string> {
+    const workspaceFolders = vscode.workspace.workspaceFolders;
+    if (workspaceFolders?.length && !path.isAbsolute(formatterUrl)) {
+        for (const workspaceFolder of workspaceFolders) {
+            const filePath = path.resolve(workspaceFolder.uri.fsPath, formatterUrl);
```
jdneo, line 13: what will happen if there are multiple workspace folders and they all have the profiles with the same name?

src/formatter-settings/utils.ts
```diff
@@ -0,0 +1,61 @@
...
+            }
+        }
+    }
+    return path.resolve(formatterUrl);
+}
+
+export async function getVSCodeSetting(setting: string, defaultValue: any): Promise<any> {
+    return await instrumentOperation("formatter.getSetting", async (operationId: string) => {
+        const config = vscode.workspace.getConfiguration(undefined, { languageId: "java" });
+        let result = config.get<any>(setting);
```
jdneo, line 25: `let result = config.get<any>(setting) ?? vscode.workspace.getConfiguration().get<any>(setting);`

src/formatter-settings/index.ts
```diff
@@ -1,99 +1,356 @@
...
+                    }
+                    this.profileElements.set(id, setting);
+                    this.profileSettings.set(id, value);
+                    this.lastElement = setting;
+                }
+                break;
+            }
+        }
+        if (!this.profileElements.size) {
+            diagnostics.push(new vscode.Diagnostic(new vscode.Range(new vscode.Position(0, 0), new vscode.Position(0, 0)), "No valid settings found in the profile."));
```
jdneo, line 212: what's the default DiagnosticSeverity

src/formatter-settings/index.ts
```diff
@@ -1,99 +1,356 @@
...
+        await javaExt.activate();
+        this.checkedRequirement = true;
+        return true;
+    }
+
+    private async initialize(document: vscode.TextDocument, webviewPanel: vscode.WebviewPanel): Promise<boolean> {
+        if (!await this.checkRequirement()) {
+            return false;
+        }
+        if (!await this.checkProfile()) {
```
jdneo, line 137: why checkProfile() is called again here?

src/formatter-settings/index.ts
CsCherrYY: The default severity is `Error`. Here we just push an `Error` since it's meaningless to use an empty profile.

src/formatter-settings/index.ts
CsCherrYY: The other `checkProfile()` is called in `showFormatterSettingsEditor()` when we execute command `java.formatterSettings`. If we open the editor in another way, such as `reload webviews` or manually open a xml using `open with...`, we need to check the profile.

src/formatter-settings/utils.ts
CsCherrYY: Currently, the java formatter form vscode.java supports only one valid profile, which comes from the first workspace. If there are multiple workspace folders, the settings.json will tell you the setting might not be applied.
workspace

src/formatter-settings/index.ts
```diff
@@ -1,99 +1,356 @@
...
+                "Yes", "No").then(async (result) => {
+                    if (result === "Yes") {
+                        vscode.commands.executeCommand("workbench.action.webview.reloadWebviewAction");
+                    }
+                });
+        }
+    }
+
+    private updateProfileSettings(document: vscode.TextDocument, webviewPanel: vscode.WebviewPanel): vscode.Diagnostic[] {
+        this.profileElements.clear();
```
jdneo, line 171: Please consider using debounce to reduce the unnecessary requests

src/formatter-settings/index.ts
```diff
@@ -1,99 +1,356 @@
...
+        this.supportedProfileSettings = getSupportedProfileSettings(Number(this.settingsVersion));
+        for (const setting of this.supportedProfileSettings.values()) {
+            const element = this.profileElements.get(setting.id);
+            const value = this.profileSettings.get(setting.id);
+            if (!element || !value) {
+                continue;
+            }
+            const webViewValue: string | undefined = FormatterConverter.profile2WebViewConvert(setting.id, value);
+            if (!webViewValue) {
+                const elementContent = new XMLSerializer().serializeToString(element);
```
jdneo, line 224: There will be some errors in the console if the xml is corrupted

src/formatter-settings/index.ts
```diff
@@ -1,99 +1,356 @@
...
+                    }
+                });
+        }
+    }
+
+    private updateProfileSettings(document: vscode.TextDocument, webviewPanel: vscode.WebviewPanel): vscode.Diagnostic[] {
+        this.profileElements.clear();
+        this.profileSettings.clear();
+        this.lastElement = undefined;
+        this.diagnosticCollection.clear();
```
jdneo, line 174: If you only update the diagnostics when the custom editor is opened, after closing it, these diagnostics will be always there.

src/formatter-settings/utils.ts
```diff
@@ -0,0 +1,149 @@
...
+}
+
+export function parseProfile(document: vscode.TextDocument): ProfileContent {
+    const profileElements = new Map<string, DOMElement>();
+    const profileSettings = new Map<string, string>();
+    let lastElement = undefined;
+    const diagnostics: vscode.Diagnostic[] = [];
+    const documentDOM = new DOMParser({
+        locator: {}, errorHandler: (_level, msg) => {
+            const bracketExp: RegExp = new RegExp("\\[(.*?)\\]", "g");
```
CsCherrYY, line 71: DOM parse error message example:
`[xmldom error]element parse error: Error: attribute value must after "="\n@#[line:17,col:9]`

src/formatter-settings/assets/features/formatterSettings/components/Setting.tsx
```diff
@@ -30,6 +30,7 @@ const Setting = (): JSX.Element => {
     if (!value || value === "") {
       value = "0";
     }
+    value = Number(value);
```
jdneo, line 33: I think the value can directly pass to the server side

src/formatter-settings/index.ts
```diff
@@ -1,99 +1,306 @@
...
+            edit.replace(document.uri, new vscode.Range(new vscode.Position(profileElement.lineNumber - 1, profileElement.columnNumber - 1), new vscode.Position(profileElement.lineNumber - 1, profileElement.columnNumber - 1 + originalSetting.length)), new XMLSerializer().serializeToString(profileElement));
+            await vscode.workspace.applyEdit(edit);
+        }
+    }
+
+    private async checkProfile(): Promise<boolean> {
+        if (this.checkedProfile) {
+            return true;
+        }
+        this.checkedProfile = await instrumentOperation("formatter.checkProfileSetting", async (operationId: string) => {
```
jdneo, line 254: why not assign instrumentOperation to a function

src/formatter-settings/index.ts
```diff
@@ -1,99 +1,306 @@
...
+                    command: "loadProfileSetting",
+                    setting: Array.from(content.supportedProfileSettings.values()),
+                });
+            }
+            await this.updateVSCodeSettings();
+            this.format();
+        }
+    }
+
+    private onChangeProfile(): void {
```
jdneo, line 180: We need to distinguish profile and the setting pointing to the profile location.

`onChangeProfile()` sounds like the event when the profile file is changed

src/formatter-settings/index.ts
```diff
@@ -1,99 +1,306 @@
...
+    private lastElement: DOMElement | undefined;
+    private settingsVersion: string = JavaConstants.CURRENT_FORMATTER_SETTINGS_VERSION;
+    private checkedRequirement: boolean = false;
+    private checkedProfile: boolean = false;
+    private diagnosticCollection: vscode.DiagnosticCollection = vscode.languages.createDiagnosticCollection();
+    private settingsUrl: string | undefined = vscode.workspace.getConfiguration("java").get<string>(JavaConstants.SETTINGS_URL_KEY);
+    private webviewPanel: vscode.WebviewPanel | undefined;
 
     constructor(private readonly context: vscode.ExtensionContext) {
+        vscode.workspace.onDidChangeConfiguration(e => {
```
jdneo, line 30: Two of this listener?

src/formatter-settings/index.ts
```diff
@@ -1,99 +1,306 @@
...
+            if (e.affectsConfiguration(`java.${JavaConstants.SETTINGS_URL_KEY}`)) {
+                this.checkedProfile = false;
+                this.settingsUrl = vscode.workspace.getConfiguration("java").get<string>(JavaConstants.SETTINGS_URL_KEY);
+            } else if (e.affectsConfiguration(`java.${JavaConstants.SETTINGS_PROFILE_KEY}`)) {
+                this.checkedProfile = false;
+            }
+        });
+        vscode.workspace.onDidChangeTextDocument(async (e: vscode.TextDocumentChangeEvent) => {
+            if (!this.settingsUrl || e.document.uri.toString() !== vscode.Uri.file(await getProfilePath(this.settingsUrl)).toString()) {
```
jdneo, line 39: Can we cache the location of the profile and update it when setting changes?

src/formatter-settings/utils.ts
```diff
@@ -0,0 +1,149 @@
...
+        const value = profileSettings.get(setting.id);
+        if (!element || !value) {
+            setting.value = FormatterConverter.profile2WebViewConvert(setting.id, getDefaultValue(setting.id))!;
+            continue;
+        }
+        const webViewValue: string | undefined = FormatterConverter.profile2WebViewConvert(setting.id, value);
+        if (!webViewValue) {
+            const elementContent = new XMLSerializer().serializeToString(element);
+            const elementRange = new vscode.Range(new vscode.Position(element.lineNumber - 1, element.columnNumber - 1), new vscode.Position(element.lineNumber - 1, element.columnNumber - 1 + elementContent.length));
+            diagnostics.push(new vscode.Diagnostic(elementRange, `Invalid value in id: "${setting.id}", "${value}" is not supported.`, vscode.DiagnosticSeverity.Error));
```
jdneo, line 134: More accurate position

src/formatter-settings/utils.ts
```diff
@@ -0,0 +1,149 @@
...
+            profileSettings.delete(setting.id);
+            setting.value = FormatterConverter.profile2WebViewConvert(setting.id, getDefaultValue(setting.id))!;
+            continue;
+        }
+        setting.value = webViewValue;
+    }
+    return {
```
jdneo, line 141:
```js
return {
    profileElements,
    profileSettings,
    lastElement,
    supportedProfileSettings,
    settingsVersion,
    diagnostics
}

File: src/formatter-settings/utils.ts (hunk 8 above)
Comment (jdneo) [reply src/formatter-settings/utils.ts:71]: If the output format is stable, why you do not get all the needed fields in just one regexp?

File: src/formatter-settings/index.ts (hunk 11 above)
Comment (CsCherrYY) [reply src/formatter-settings/index.ts:180]: will change the ambiguity names.

File: src/formatter-settings/index.ts (hunk 12 above)
Comment (CsCherrYY) [reply src/formatter-settings/index.ts:30]: Will merge them into one listener

File: src/formatter-settings/index.ts [hunk 16]
```diff
@@ -39,7 +38,8 @@ export class JavaFormatterSettingsEditorProvider implements vscode.CustomTextEdi
             if (e.affectsConfiguration(`java.${JavaConstants.SETTINGS_URL_KEY}`)) {
                 this.settingsUrl = vscode.workspace.getConfiguration("java").get<string>(JavaConstants.SETTINGS_URL_KEY);
                 if (this.settingsUrl) {
-                    this.profilePath = await getProfilePath(this.settingsUrl);
+                    const res = await getProfilePath(this.settingsUrl);
```
jdneo, line 41: What's the purpose to introduce the variable `res` here?

src/formatter-settings/utils.ts
```diff
@@ -37,8 +37,9 @@ export function isRemote(path: string): boolean {
     return path !== null && path.startsWith("http:/") || path.startsWith("https:/");
 }
 
-export async function getTargetProfilePath(context: vscode.ExtensionContext, fileName?: string): Promise<string> {
-    const targetFileName = fileName || "java-formatter.xml";
+export async function addDefaultProfile(context: vscode.ExtensionContext): Promise<void> {
+    const defaultProfile: string = path.join(context.extensionPath, "webview-resources", "java-formatter.xml");
+    const targetFileName = "java-formatter.xml";
     let profilePath: string;
     const workspaceFolder = vscode.workspace.workspaceFolders;
```
jdneo, line 44: workspaceFolder`s`

jdneo:
The current implementation works well. But there are space to improve it. To me this can be transformed to a `MVC`/`MPV` way. Your customer editor works as a controller/presenter. There will be another provider for the setting profile, to resolve the backend model. This will decouple the components. Please consider to change the implementation when you have time

CsCherrYY:
> The current implementation works well. But there are space to improve it. To me this can be transformed to a `MVC`/`MPV` way. Your customer editor works as a controller/presenter. There will be another provider for the setting profile, to resolve the backend model. This will decouple the components. Please consider to change the implementation when you have time

Sure. I have created an engineering issue #633 to record this task.

jdneo:
New bug: the input box cannot remove by the `Backspace`

src/formatter-settings/utils.ts
```diff
@@ -39,22 +42,28 @@ export function isRemote(path: string): boolean {
...
     }
-    // bug: https://github.com/redhat-developer/vscode-java/issues/1944, only the profiles in posix path can be monitored when changes, so we use posix path for default profile creation temporarily.
-    const relativePath = toPosixPath(path.join(".vscode", targetFileName));
+     // bug: https://github.com/redhat-developer/vscode-java/issues/1944, only the profiles in posix path can be monitored when changes, so we use posix path for default profile creation temporarily.
+    const relativePath = toPosixPath(path.join(".vscode", fileName));
     profilePath = toPosixPath(profilePath);
-    await fse.copy(defaultProfile, profilePath);
-    await vscode.workspace.getConfiguration("java").update("format.settings.url", (workspaceFolders?.length ? relativePath : profilePath), !(workspaceFolders?.length));
-    vscode.commands.executeCommand("vscode.openWith", vscode.Uri.file(profilePath), "java.formatterSettingsEditor");
+    return {relativePath, profilePath};
```
jdneo, line 66: Can you use `workspace.asRelativePath()` to get the relative path instead of returning a object?

src/formatter-settings/utils.ts
```diff
@@ -148,3 +157,65 @@ export function parseProfile(document: vscode.TextDocument): ProfileContent {
...
 }
+
+export async function downloadFile(settingsUrl: string, extensionVersion: string, targetPath?: string): Promise<string> {
+    if (targetPath) {
+        await fse.ensureDir(path.dirname(targetPath));
+        if (await fse.pathExists(targetPath)) {
+            await fse.remove(targetPath);
+        }
+    }
+    return await new Promise((resolve: (res: string) => void, reject: (e: Error) => void): void => {
```
jdneo, line 168: Can we use some third-party library to download the file, like `Axios`

CsCherrYY: Sure. Will try `Axios` then.

jdneo (approved):

PR #643, opened 2021-05-24
fix: Support Open Settings shortcut in the editor title area

### LINKED ISSUE #636: [Formatting Preview] Support Open Settings shortcut in the editor title area
Like the native VS Code setting

CsCherrYY:
fix #636

### DIFF
--- a/package.json
+++ b/package.json
@@ -199,6 +199,12 @@
         "command": "java.formatterSettings",
         "category": "Java",
         "title": "Open Java Formatter Settings with Preview"
+      },
+      {
+        "command": "java.formatterSettings.showTextEditor",
+        "category": "Java",
+        "title": "Open Text Editor",
+        "icon": "$(go-to-file)"
       }
     ],
     "configuration": {
@@ -252,6 +258,13 @@
           "command": "java.classpathConfiguration",
           "group": "9_configuration@20",
           "when": "view == javaProjectExplorer && viewItem =~ /java:project(?=.*?\\b\\+unmanagedFolder\\b)(?=.*?\\b\\+uri\\b)/"
+        }
+      ],
+      "editor/title": [
+        {
+          "command": "java.formatterSettings.showTextEditor",
+          "when": "activeCustomEditorId == java.formatterSettingsEditor && !readOnlyMode",
+          "group": "navigation"
         }
       ]
     },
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -253,6 +259,13 @@
           "group": "9_configuration@20",
           "when": "view == javaProjectExplorer && viewItem =~ /java:project(?=.*?\\b\\+unmanagedFolder\\b)(?=.*?\\b\\+uri\\b)/"
         }
+      ],
+      "editor/title": [
+        {
+          "command": "java.formatterSettings.showTextEditor",
+          "when": "activeCustomEditorId == java.formatterSettingsEditor && !readOnlyMode",
```
jdneo, line 266: I think it's ok to show the readonly text editor when the setting points to a url

package.json
```diff
@@ -199,6 +199,12 @@
         "command": "java.formatterSettings",
         "category": "Java",
         "title": "Open Java Formatter Settings with Preview"
+      },
+      {
+        "command": "java.formatterSettings.showTextEditor",
```
jdneo, line 204: disable it in other unnecessary entries, like the command palette

src/formatter-settings/index.ts
```diff
@@ -61,6 +62,14 @@ export class JavaFormatterSettingsEditorProvider implements vscode.CustomTextEdi
         vscode.commands.executeCommand("vscode.openWith", filePath, "java.formatterSettingsEditor");
     }
 
+    public reopenWithTextEditor = instrumentOperation("formatter.showTextEditor", (operationId: string, uri: any) => {
+        sendInfo(operationId, { editor: "RawTextEditor" });
+        if (uri instanceof vscode.Uri) {
+            this.webviewPanel?.dispose();
```
jdneo, line 68: What will happen if we do not dispose it?

CsCherrYY: We'll have another raw text editor open, without closing the current custom editor.

package.json
CsCherrYY: +1 for VSCode will show readonly tips
remote

src/formatter-settings/index.ts
CsCherrYY: Update: VSCode behavior is to open another raw editor without closing the current editor, we'd better follow VSCode's.

jdneo (approved):

PR #650, opened 2021-05-25
fix: If a remote profile failed to open, the user cannot open it again

### LINKED ISSUE #642: [Formatting Preview] If a remote profile failed to open by some network issue, the user cannot open it again
If a remote profile failed to open by some network issue, the user cannot open it again.

btw, the error message can be better as well, instead of simply throw out the raw error message

CsCherrYY:
fix #642

### DIFF
--- a/src/formatter-settings/utils.ts
+++ b/src/formatter-settings/utils.ts
@@ -156,9 +156,8 @@ export function parseProfile(document: vscode.TextDocument): ProfileContent {
 }
 
 export async function downloadFile(settingsUrl: string): Promise<string> {
-    let content = "";
     try {
-        content = (await axios.get(settingsUrl)).data;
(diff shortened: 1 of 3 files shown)


src/formatter-settings/utils.ts
```diff
@@ -156,7 +156,18 @@ export function parseProfile(document: vscode.TextDocument): ProfileContent {
 }
 
 export async function downloadFile(settingsUrl: string): Promise<string> {
-    return (await axios.get(settingsUrl)).data;
+    let content = "";
+    try {
+        content = (await axios.get(settingsUrl)).data;
```
jdneo, line 161: directly return `(await axios.get(settingsUrl)).data;`

jdneo (approved):

Issue #654, opened 2021-05-27
JAVA DEPENDENCIES view not available


Kayes-Islam:
**Environment:**

- VS Code Version: 1.56.2 (user setup)
- Extension Version: v0.15.0
- JDK Type & Version: java version "15.0.2" 2021-01-19, Java(TM) SE Runtime Environment (build 15.0.2+7-27), Java HotSpot(TM) 64-Bit Server VM (build 15.0.2+7-27, mixed mode, sharing)
- OS Type & Version: Windows 10

**Step to Reproduce:**
1. On an existing VS Code setup, install the extension pack following this links: https://code.visualstudio.com/docs/languages/java
2. Opened the command pallet and select "Crete Java Projec..."
3. Select "No build tools" 
4. Select folder path, project names etc.
5. Open the readme.md which mentions about JAVA DEPENDENCIES view. Open the link provided. Notice that the image link is broken. 
6. No JAVA DEPENDENCIES view available.
6. Open a java file. Still no JAVA DEPENDENCIES view available.

![image](https://user-images.githubusercontent.com/9096105/119746412-99a27280-becf-11eb-94d6-a13c66c55f92.png)

**Expected Behavior:**
1. JAVA DEPENDENCIES view is available for a java project.
2. The linked documentation page should tell me what it looks like. (image link is broken at the moment).
**Actual Behavior:**

**Related Issues:**

CsCherrYY:
The JAVA DEPENDENCIES view has been renamed to JAVA PROJECTS view and you can find it in your screenshot. 

We'll update the README to synchronize.

Kayes-Islam:
Thanks @CsCherrYY, I'm from .Net coming to Java after a long time so had no clue :)

The readme says:
>The `JAVA DEPENDENCIES` view allows you to manage your dependencies. More details can be found [here](https://github.com/microsoft/vscode-java-pack/blob/master/release-notes/v0.9.0.md#work-with-jar-files-directly).

Reading that I was expecting to have some NuGet/npm like dependency management functionality for allowing me to add/remove packages on my project, without it being a Maven project. However I did now see any such functionality available on the view or when I right click on any of the nodes on the view. What am I missing here?

CsCherrYY:
Do you want to add a jar file as your project dependency? You can hover on the `Referenced Libraries` node and see a `+` on the right.
[...]

PR #659, opened 2021-05-31
fix: Release-notes are broken

CsCherrYY:
See: #654 

The screenshots of release-notes are still using `master` branch name although we have renamed the branch to `main`, it makes the screenshots broken. 

Since `reference-jar-files.gif` and related section will be found when creating new Java project in Project Manager and the current note is using the old view name, we'll update it to avoid misunderstandings.


release-notes/v0.9.0.md
```diff
@@ -135,7 +135,7 @@ We’d like to encourage more developers to join us and improve Java development
 
 ## Work with JAR Files Directly
 
-If you just started to learn Java and don't have any clue what Maven or Gradle is, we got you back. Now VS Code Java lets you work with JAR files directly without any build tools. Go to `JAVA DEPENDENCIES` view, find the `Referenced Libraries` node and click the `+` icon:
```
Eskibear, line 138: I suggest not to change the name. Because at that time (v0.9.0), the explorer was called "JAVA DEPENDENCIES". It's wired to change content of historical release notes.

CsCherrYY: How about adding a note to show that we're using `JAVA PROJECTS` since a specific version?

Eskibear (approved):

PR #660, opened 2021-05-31
update latest JDK to 16

Eskibear:


Eskibear:
In long term, it should be fetched from metadata, tracked in #661

testforstephen (approved):

PR #664, opened 2021-05-31
fix: "formatterExample" in telemetry is regarded as measure

CsCherrYY:
Currently the value of `formatterExample` is number, so it was regarded as measure. Here to fix it, and send info only when example changes.

### DIFF
--- a/src/formatter-settings/FormatterConstants.ts
+++ b/src/formatter-settings/FormatterConstants.ts
@@ -279,6 +279,33 @@ export namespace Example {
                 return "";
         }
     }
+
+    export function getExampleTitle(example: ExampleKind): string {
+        switch (example) {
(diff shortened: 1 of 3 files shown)


src/formatter-settings/FormatterConstants.ts
```diff
@@ -279,6 +279,33 @@ export namespace Example {
                 return "";
         }
     }
+
+    export function getExampleTitle(example: ExampleKind): string {
```
Eskibear, line 283: Do we have string-based enum in TS?

CsCherrYY: Yes, we can use it. I'll have a try.

Eskibear (approved):

PR #669, opened 2021-06-11
add more categories for better visibility

Eskibear:
Nick found Java Ext Pack is missing with below filtering condition, generated by VS Code's native getting started walkthrough.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,9 @@
   ],
   "categories": [
     "Programming Languages",
+    "Linters",
+    "Debuggers",
+    "Formatters",
     "Extension Packs"
   ],
   "repository": {
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "Linters",
     "Debuggers",
     "Formatters",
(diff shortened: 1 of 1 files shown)


nickzhums:
should we also add "linter" / "debugger" / "formatter" since it's part of the extension pack ? (I'm taking Python as reference: https://marketplace.visualstudio.com/items?itemName=ms-python.python)

thanks!

Eskibear:
yes we can. I don't see any side effect by adding them.

nickzhums:
> yes we can. I don't see any side effect by adding them.

awesome !

package.json
```diff
@@ -19,6 +19,10 @@
     "spring"
   ],
   "categories": [
+    "Programming Languages",
+    "Linters",
+    "Debuggers",
+    "Formatters",
```
testforstephen, line 25: Add "Snippets" as well since Java language support provides snippets.

testforstephen (approved):

PR #682, opened 2021-06-29
npm audit fix

Eskibear:


Eskibear:
There's big effort migrating webpack from 4 to 5, as lots of plugins don't support webpack 5 and are no longer maintained.

I'm closing it for the moment.

Issue #692, opened 2021-07-21
improve getting started experience with walkthrough


Eskibear:
### A/B testing hosted by VS Code team
- [x] Make sure the walkthrough page and Java Welcome page won't show at the same time. #700 

### Update walkthrough page:
- [x] The last step will navigate user to the welcome page #698 

### Update the Java Welcome page:
- [x] Change the name to 'Java Help Center' or sth else. #694 
- [x] Section 'Configuration' -> 'General'  #694 
- [x] Set to the default not show #699 

### Update Getting Started page:
- [x] Rename the title to sth like 'Coding and debugging tips for students'  #693 
- [x] Also update the title in the Student section in Java Welcome page #693 
- [x] Remove the part for 'No import needed'  #693

PR #693, opened 2021-07-21
rename student getting started page

Eskibear:
See #692

### DIFF
--- a/src/getting-started/assets/index.html
+++ b/src/getting-started/assets/index.html
@@ -4,14 +4,14 @@
 <head>
   <meta charset="UTF-8">
   <meta http-equiv="Content-Security-Policy" content="default-src 'unsafe-inline'">
-  <title>Java Getting Started</title>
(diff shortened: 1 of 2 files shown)


jdneo (dismissed):

src/getting-started/assets/index.html
```diff
@@ -4,14 +4,14 @@
 <head>
   <meta charset="UTF-8">
   <meta http-equiv="Content-Security-Policy" content="default-src 'unsafe-inline'">
-  <title>Java Getting Started</title>
+  <title>Coding and Debugging Tips for Students</title>
```
testforstephen, line 7: Change it to "for Beginners"? Since this is also useful to new VS Code user.

Eskibear: good suggestion

Eskibear: updated.

testforstephen (approved):

PR #696, opened 2021-07-23
walkthrough: add links to debug tutorial

Eskibear:
image

linked to https://code.visualstudio.com/docs/java/java-debugging

### DIFF
--- a/package.json
+++ b/package.json
@@ -126,7 +126,7 @@
           {
             "id": "java.showDebugView",
             "title": "Launch and debug your app",
-            "description": "The debugger will automatically find the entry point of your project, launch it and start debugging.\n[Show Run and Debug View](command:workbench.view.debug)",
+            "description": "The debugger will automatically find the entry point of your project, launch it and start debugging.\n[Show Run and Debug View](command:workbench.view.debug)\n[Learn More...](https://code.visualstudio.com/docs/java/java-debugging)",
             "media": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -126,7 +126,7 @@
           {
             "id": "java.showDebugView",
             "title": "Launch and debug your app",
-            "description": "The debugger will automatically find the entry point of your project, launch it and start debugging.\n[Show Run and Debug View](command:workbench.view.debug)",
+            "description": "The debugger will automatically find the entry point of your project, launch it and start debugging.\n[Show Run and Debug View](command:workbench.view.debug)\n[Learn More...](https://code.visualstudio.com/docs/java/java-debugging)",
```
jdneo, line 129: I'm thinking that if it would be helpful if trigger a command to open the link here?

Eskibear: what for? telemetry?

jdneo: yes

Eskibear: Pros:
- We'll tell how many users visit debugging docs from walkthrough (vs overview/welcome/etc.)

Cons:
- We'll have to activated the extension pack on executing that command.
- This link will look complicated and hard to maintain?

jdneo: > We'll have to activated the extension pack on executing that command.

Ok, this is a good point we should not use a command

jdneo (approved):

PR #700, opened 2021-07-23
refine welcome view for getting started experiment

Eskibear:
Change:
- For all users, first view will not show by default.

About EXP:
use `vscode.gettingStarted.overrideCategory.vscjava.vscode-java-pack.javaWelcome.when` as feature gate. @digitarald please use the same name when you set up experiment later.

- Treatment group
  - no page will be open by vscode-java-pack extension.
  - walkthrough is expected to show after installation 
- Control group
  - we will show the "welcome webbiew" only **once** when extension is activated.
  - walkthough not visible, because by default it's disabled by `when: false`.


Eskibear:
Update: 
For treatment group, "Take a tour" points to walkthrough as below:

image

jdneo (commented):
LGTM

jdneo (approved):

PR #701, opened 2021-07-26
continue #700: support local overridden by settings

Eskibear:
Before EXP is created, we use local settings to override it. 

> "experiments.override.gettingStarted.overrideCategory.vscjava.vscode-java-pack#javaWelcome.when": "true",

### DIFF
--- a/src/utils/walkthrough.ts
+++ b/src/utils/walkthrough.ts
@@ -1,0 +1,15 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+import { workspace } from "vscode"; 
+import { getExpService } from "../exp";
+import { TreatmentVariables } from "../exp/TreatmentVariables";
+
+export function isWalkthroughEnabled() {
+
+    const fromExp = getExpService().getTreatmentVariable<boolean>(TreatmentVariables.VSCodeConfig, TreatmentVariables.JavaWalkthroughEnabled);
+    // 	can be overridden by local settings "experiments.override.gettingStarted.overrideCategory.vscjava.vscode-java-pack#javaWelcome.when": "true"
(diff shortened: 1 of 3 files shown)


src/utils/walkthrough.ts
```diff
@@ -0,0 +1,15 @@
...
+import { getExpService } from "../exp";
+import { TreatmentVariables } from "../exp/TreatmentVariables";
+
+export function isWalkthroughEnabled() {
+
+    const fromExp = getExpService().getTreatmentVariable<boolean>(TreatmentVariables.VSCodeConfig, TreatmentVariables.JavaWalkthroughEnabled);
+    // 	can be overridden by local settings "experiments.override.gettingStarted.overrideCategory.vscjava.vscode-java-pack#javaWelcome.when": "true"
+    const fromSettings = workspace.getConfiguration("experiments.override.gettingStarted.overrideCategory").get<string>("vscjava.vscode-java-pack#javaWelcome.when") === "true";
+
+    return fromExp || fromSettings;
```
jdneo, line 14: what if user explicitly set setting to false and the value from exp service is true?

Eskibear: It's true. No one is supposed to do that. 

 `experiments.override.gettingStarted` is designed for early dogfood the walkthrough, while don't affect most users.

Eskibear: Or do you want me to weight `experiments.override.gettingStarted` up? So that we can test the behavior no matter which group we are in. 

Sound good idea, but I don't know the behavior of vscode, whether it enables walkthrough in that case.

jdneo: I think most users might not aware of this setting. We don't expose that in the UI right?

Eskibear: No, it's not exposed. It's a hidden setting.

jdneo: Ok then this should not be a problem I guess

jdneo (approved):

PR #704, opened 2021-07-29
upversion to 0.18.1

Eskibear:
For those who are used to "Java: Welcome", in case they cannot find the new command, we append "(Welcome)" after "Help Center".


jdneo (approved):

Eskibear:
@nickzhums please take a look, see if it's ok?

PR #713, opened 2021-08-02
fix: tour page not shown at first launch

Eskibear:


Eskibear:
After open welcome page, this state will be update automatically. 
https://github.com/microsoft/vscode-java-pack/blob/35b45abf6014fdc04abc9f6582956c9f2d4c1c66/src/welcome/index.ts#L99

jdneo (approved):

PR #723, opened 2021-08-10
open help center to side on walkthrough

Eskibear:

### DIFF
--- a/src/welcome/index.ts
+++ b/src/welcome/index.ts
@@ -34,7 +34,6 @@ export async function showWelcomeWebview(context: vscode.ExtensionContext, _oper
         welcomeView = vscode.window.createWebviewPanel(
             "java.welcome",
             "Java Help Center",
-            // vscode.ViewColumn.Beside,
             viewColumn,
             {
                 enableScripts: true,
--- a/src/welcome/index.ts
+++ b/src/welcome/index.ts
@@ -10,7 +10,18 @@ import { isWalkthroughEnabled } from "../utils/walkthrough";
 
 let welcomeView: vscode.WebviewPanel | undefined;
 
-export async function showWelcomeWebview(context: vscode.ExtensionContext, _operationId?: string, options?: any) {
+// from command palette
+export async function showWelcomeWebviewBeside(context: vscode.ExtensionContext, _operationId?: string, options?: {
+    firstTimeRun?: boolean;
+}) {
+    const newOptions = { openBeside: true, firstTimeRun: options?.firstTimeRun };
+    await showWelcomeWebview(context, _operationId, newOptions);
+}
+
+export async function showWelcomeWebview(context: vscode.ExtensionContext, _operationId?: string, options?: {
+    firstTimeRun?: boolean;
+    openBeside?: boolean;
+}) {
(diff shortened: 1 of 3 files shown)


Eskibear:
### Treatment Group: 

https://user-images.githubusercontent.com/2351748/128839239-71be8c91-91bc-4fd6-8f23-cd05884a9c0d.mp4



### Control Group:


https://user-images.githubusercontent.com/2351748/128839252-a3bf4ec2-64ab-4d3c-8a14-ce24974822a7.mp4

src/welcome/index.ts
```diff
@@ -10,7 +10,18 @@ import { isWalkthroughEnabled } from "../utils/walkthrough";
 
 let welcomeView: vscode.WebviewPanel | undefined;
 
-export async function showWelcomeWebview(context: vscode.ExtensionContext, _operationId?: string, options?: any) {
+// from command palette
+export async function showWelcomeWebviewBeside(context: vscode.ExtensionContext, _operationId?: string, options?: {
+    firstTimeRun?: boolean;
+}) {
+    const newOptions = { openBeside: true, firstTimeRun: options?.firstTimeRun };
```
jdneo, line 17: Directly use viewColumn as the option?

`{ viewColumn: vscode.ViewColumn.Beside, firstTimeRun: options?.firstTimeRun }`

src/welcome/index.ts
```diff
@@ -19,10 +30,12 @@ export async function showWelcomeWebview(context: vscode.ExtensionContext, _oper
         welcomeView.reveal();
         fetchInitProps(context);
     } else {
+        const viewColumn = options?.openBeside ? vscode.ViewColumn.Beside : vscode.ViewColumn.Active;
         welcomeView = vscode.window.createWebviewPanel(
             "java.welcome",
             "Java Help Center",
-            vscode.ViewColumn.Beside,
+            // vscode.ViewColumn.Beside,
```
jdneo, line 37: remove?

src/welcome/index.ts
Eskibear: > Directly use viewColumn as the option?

That's would be unstraightforward if triggered from url... E.g. previous `command://java.welcome?openBiside=true` would become `command://java.welcome?viewColumn=-2`. (because current value of the enum `ViewColumn.Beside` is `-2`.

jdneo: Hmm, honestly I didn't see any problem with `java.welcome?viewColumn=-2`. Maybe in the future we want to have a new different position? Or VS Code proposed a new position called `Bottom`, who knows.

Anyway, not a big problem, leave the choice to you.

jdneo (approved):

src/welcome/index.ts
Eskibear: > in the future we want to have a new different position? Or VS Code proposed a new position called Bottom

I don't think vscode teams will easily change value of enum (though there's possibility). My point is, for future maintainers `openBeside=true` is more readable comparing with `viewColumn=-2`.

jdneo: Not change but add. Like, a new value `viewColumn.Bottom` and user request that

PR #736, opened 2021-09-06
Add lombok to readme

nickzhums:


nickzhums:
ping

Eskibear:
Sorry for delayed response, below are just my two cents:

Good points:
* Lombok is a popular tool. Adding this will make the community know that vscode has a extension supporting lombok.

My concerns:
- So far I've seen multiple issues related to gabrielbb.vscode-lombok, see https://github.com/redhat-developer/vscode-java/issues?q=is%3Aissue+gabrielbb+ , sometimes it prevents LS from launching. I'm wondering if these bugs have already been addressed. 
- Since Java pack is not including this extension, is it good to add it in README? It's a little bit confusing.

nickzhums:
thanks! @Eskibear 

Your concerns are all valid, but I do want to provide directions for developer who are using different frameworks.

What do you think about we add a section at the end of this README, and name it "Support of other frameworks", and add description such as "Those extensions are not maintained by Microsoft so use them at your own discretion", then list extensions such as Lombok?

testforstephen:
We saw some issues when install/uninstall lombok extension, for example, it's possible to cause LS to fail to start. And that extension doesn't seem to be maintained anymore. I prefer to hold on the recommendation.

nickzhums:
Got it, let's hold off for Lombak now - thanks guys. I just want to provide more guidance for the user when they are looking for extensions use, because we won't be able to support every framework - any ideas? Maybe we can use the extensions guide in our Help Center

nickzhums:
Based on the discussion, we decide not to include this for now, so closing this PR

PR #732, opened 2021-08-30
clarify name of extensions

Eskibear:
See https://github.com/microsoft/vscode-java-pack/issues/729#issuecomment-907974405

PR #738, opened 2021-09-07
Normalize extension naming

### LINKED ISSUE #729: Ambiguous references to "Java extension"
Although this extension pack doesn't include any extension that calls itself "Java extension" (nor does the pack calls itself that), there are several user-facing messages that make a reference to that expression. For example:

https://github.com/microsoft/vscode-java-pack/blob/378b409d49032b03cbf0032cc51a6031009e0829/src/java-runtime/utils/upstreamApi.ts#L52

https://github.com/microsoft/vscode-java-pack/blob/378b409d49032b03cbf0032cc51a6031009e0829/src/java-runtime/index.ts#L209

https://github.com/microsoft/vscode-java-pack/blob/a5f9421a3aedb13afa5caa171eddcc5a9d9a1f27/src/getting-started/assets/index.html#L326

From what I can infer, "Java extension" is a reference to the [**Language Support for Java by Red Hat**](https://marketplace.visualstudio.com/items?itemName=redhat.java) extension. Indeed, there are places where the extension is referred to explicitly by its slug `redhat.java`:

https://github.com/microsoft/vscode-java-pack/blob/e61dd48f623f42220af22aa4adfe8746f24aa839/src/formatter-settings/index.ts#L164-L166

I realize the full name of the extension is kind of a mouthful, but there needs to be some disambiguation in these messages, since there's no explicit reference to that extension being "the" Java extension (even though, for all practical purposes, it is).

What would be the preferred way to resolve this? Actually use the full name in the messages linked above? Or is there a shortened form that would be acceptable (e.g. "RedHat Java extension")?

I'd be happy to submit a PR making these changes once there's consensus on how to proceed.

waldyrious:
- Tweak name of the Java Extensions Guide, so it doesn't mix up with either the redhat.java extension or the Java extension pack;  
- Minor phrasing corrections (mostly adding articles where needed);
- Normalize the name of the redhat.java extension

Fixes #729. Replaces #732.

### DIFF
--- a/package.json
+++ b/package.json
@@ -212,7 +212,7 @@
       {
         "command": "java.extGuide",
         "category": "Java",
-        "title": "Extension Guide"
+        "title": "Extensions Guide"
       },
       {
         "command": "java.formatterSettings",
--- a/src/classpath/assets/features/classpathConfiguration/components/Exception.tsx
+++ b/src/classpath/assets/features/classpathConfiguration/components/Exception.tsx
@@ -26,14 +26,14 @@ const Exception = (): JSX.Element | null => {
     case ClasspathViewException.JavaExtensionNotInstalled:
       content = (
         <div>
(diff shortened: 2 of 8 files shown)


Eskibear:
LGTM. invite @nickzhums to review.

nickzhums (approved):

Issue #768, opened 2021-10-14
missing third party notice


Eskibear:
Extensions missing 3rd party notice:
- [ ] pack @Eskibear 
- [ ] test runner @jdneo 
- [ ] project manager @jdneo 
- [ ] debugger (2 repos) @testforstephen 
- [ ] maven @Eskibear 
- [ ] spring-initializr @Eskibear 
- [ ] spring-boot-dashboard @Eskibear 
- [ ] gradle @CsCherrYY 

Action to take:
* add `ThirdPartyNotices.txt` containing all licenses of dependencies.

Eskibear:
another tool parsing TPN from js.map: https://www.npmjs.com/package/@rnx-kit/third-party-notices

PR #784, opened 2021-10-19
update third party notices

Eskibear:
part of microsoft/vscode-java-pack#768


jdneo:
Some used dependencies are listed as dev-dependency(i.e. react, redux, bootstrap), I think they should also be contained in the notice?

Eskibear:
> But in the production bits they are actually contained

That would be strange if code from devDependencies are contained in production bits...

PR #800, opened 2021-11-04
Eliminate high-sev CVEs by using Webpack@5

Eskibear:
This PR is a trial to adopt Webpack@5.

Here I upgrade some libs, to fix all CVEs with `high serverity`. BUT, it breaks the webviews, as html-inline plugin is no longer working with Webpack@5. 

@CsCherrYY FYI.


webpack.config.js
```diff
@@ -66,46 +75,42 @@ module.exports = function (env, argv) {
...
       }),
       new HtmlWebpackPlugin({
         filename: path.resolve(__dirname, 'out/assets/formatter-settings/index.html'),
         template: 'src/formatter-settings/assets/index.html',
-        inlineSource: '.(js|css)$',
         chunks: ['formatter-settings']
       }),
-      new HtmlWebpackInlineSourcePlugin(),
+      new InlineChunkHtmlPlugin(HtmlWebpackPlugin, [/\.(js)$/]),
+      new webpack.ProvidePlugin({
```
CsCherrYY, line 111: In webpack 5 automatic node.js polyfills are removed. Please refer to https://github.com/webpack/changelog-v5#automatic-nodejs-polyfills-removed and https://stackoverflow.com/questions/65018431/webpack-5-uncaught-referenceerror-process-is-not-defined

webpack.config.js
```diff
@@ -66,46 +75,42 @@ module.exports = function (env, argv) {
...
         chunks: ['classpath']
       }),
       new HtmlWebpackPlugin({
         filename: path.resolve(__dirname, 'out/assets/formatter-settings/index.html'),
         template: 'src/formatter-settings/assets/index.html',
-        inlineSource: '.(js|css)$',
         chunks: ['formatter-settings']
       }),
-      new HtmlWebpackInlineSourcePlugin(),
+      new InlineChunkHtmlPlugin(HtmlWebpackPlugin, [/\.(js)$/]),
```
CsCherrYY, line 110: Here we use react-dev-utils/InlineChunkHtmlPlugin for an alternative one since `html-webpack-inline-source-plugin` doesn't support webpack 5. 

Refer to https://github.com/facebook/create-react-app/tree/main/packages/react-dev-utils#new-inlinechunkhtmlpluginhtmlwebpackplugin-htmlwebpackplugin-tests-regex, it works well with HtmlWebpackPlugin 4.x, so we use `4.5.0` of `html-webpack-plugin` and it's corresponding webpack@5. Not sure the version can resolves these CVEs or not.

CsCherrYY:
currently the newest version of `react-dev-utils` (11.0.4) includes some vulnerability issues and there is an issue tracking. See: https://github.com/facebook/create-react-app/issues/11539. Let's keep eyes on it and update this PR.

Eskibear (commented):
Great job!

webpack.config.js
```diff
@@ -47,11 +47,12 @@ module.exports = function (env, argv) {
         }]
       }, {
         test: /\.(jpg|png|svg|ico|icns)$/,
-        loader: 'url-loader',
-        options: {
-          limit: 200000,
-          esModule: false
-        }
+        type: 'asset',
```
CsCherrYY, line 50: Here to remove deprecated `url-loader`, the corresponding config can be found at https://webpack.js.org/guides/asset-modules/#general-asset-type

jdneo:
Played the classpath configuration page with this PR and it works well.

webpack.config.js
Eskibear: ok, so please play around it with your webviews. If everything is working well, feel free to approve this PR and we can merge it. (I cannot approve it as original author...)

CsCherrYY (approved):
Played the formatter preview and it worked well.

PR #808, opened 2021-11-12
support to detect arm-macOS

Eskibear:
fix https://github.com/redhat-developer/vscode-java/issues/2201

### DIFF
--- a/src/utils/adoptiumApi.ts
+++ b/src/utils/adoptiumApi.ts
@@ -2,7 +2,6 @@
 // Licensed under the MIT license.
 
 import * as request from "request-promise-native";
-import architecture = require("arch");
 
 /**
  * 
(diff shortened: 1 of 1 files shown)


src/utils/adoptiumApi.ts
```diff
@@ -2,7 +2,6 @@
 // Licensed under the MIT license.
 
 import * as request from "request-promise-native";
-import architecture = require("arch");
```
jdneo, line 5: Is it safe to remove this dependency since we are not using it now?

testforstephen (approved):

src/utils/adoptiumApi.ts
Eskibear: `arch` is supposed to return actual arch of your system, instead of nodejs runtime. 
However it's no longer maintained, and returning wrong result for some platforms  like macOS M1.

The reason we introduced it was, for a x64 Windows machine, if user is running x32 VS Code, we wanted to recommend a x64 JDK, because process.arch=x32 in this case. 

But now after discussion, we find that's unnecessary. If vscode is running in x32 mode, we simply recommend a x32 JDK, which is still going tto work. VS Code also use the same way to judge architecture. And the benefit is: more consistency, no extra lib extra risk.

jdneo (approved):

PR #835, opened 2021-12-01
Fix compile warning about `debug` in `follow-redirects`

CsCherrYY:
fix 
```
WARNING in ./node_modules/follow-redirects/debug.js 7:14-30
Module not found: Error: Can't resolve 'debug' in 'C:\Users\yanzh\Work\vscode-java-pack\node_modules\follow-redirects'
Did you mean './debug'?
Requests that should resolve in the current directory need to start with './'.
```

It's a change in the webpack configuration, so I suggest a check from the page owners. @Eskibear @jdneo

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -97,7 +97,8 @@ module.exports = function (env, argv) {
     resolve: {
       modules: ['node_modules', path.resolve(__dirname, 'src')],
       mainFiles: ['index'],
-      extensions: ['.js', '.ts', '.json']
+      extensions: ['.js', '.ts', '.json'],
+      preferRelative: true
     },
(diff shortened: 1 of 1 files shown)


Eskibear:
what's effect of `preferRelative: true`?

CsCherrYY:
when resolving `require()`, the default behavior is to find the file in `node_modules`, which is the root cause of this issue. This setting change the behavior to find the file from (relative) path. In this case `./debug` is to require the `debug.js` from the same folder.

Eskibear (approved):

Eskibear:
OK, I'll create an RC bits and invite you guys to verify.

Issue #787, opened 2021-10-21
Keep the runtime requirement check consistent with the RedHat Java extension


testforstephen:
This is to adjust the UX to better support the embedded JRE in RedHat Java extension.

PR #851, opened 2021-12-16
update tooling jdk validation mechanism

Eskibear:
udpate `upstreamApi.ts` to align with changes in redhat.java extension.

The "configure java runtime" page pops up only when an invalid `java.home` is set for **generic** redhat.java extension. For platform-specific extension, tooling JDK is embedded, so we don't pop up any page.

### DIFF
--- a/src/java-runtime/utils/upstreamApi.ts
+++ b/src/java-runtime/utils/upstreamApi.ts
@@ -1,33 +1,36 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 
-// based on https://github.com/redhat-developer/vscode-java/blob/fcbe9204638610ce773cedaef445f19032a785cb/src/requirements.ts
+// based on https://github.com/redhat-developer/vscode-java/blob/c4cdbed1190fc705c364179dab525645acf03898/src/requirements.ts
 
 import * as fse from "fs-extra";
+import { findRuntimes, getRuntime, getSources, IJavaRuntime, JAVAC_FILENAME, JAVA_FILENAME } from 'jdk-utils';
 import * as path from "path";
 import * as vscode from "vscode";
-import { env } from "vscode";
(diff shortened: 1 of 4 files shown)


Eskibear:
to fix #787

src/java-runtime/utils/upstreamApi.ts
```diff
@@ -36,36 +39,117 @@ export async function resolveRequirements(): Promise<any> {
...
+                }
+            } else { // pick a default project JDK/JRE
+                /**
+                 * For legacy users, we implicitly following the order below to
+                 * set a default project JDK during initialization:
+                 * java.home > env.JDK_HOME > env.JAVA_HOME > env.PATH
+                 *
+                 * We'll keep it for compatibility.
+                 */
+                if (javaHome && (await getRuntime(javaHome) !== undefined)) {
```
CsCherrYY, line 67: Can we directly call `await getRuntime(javaHome, {withVersion: true})` only once here and store the return value?

Eskibear: yes we can, but `getRuntime` is cheap. Here I simply copy&paste to use the same logic in redhat.java.

testforstephen (approved):

src/java-runtime/utils/upstreamApi.ts
Eskibear: We can optimize it on both here and upstream later, if necessary.
e.g.

```ts
const runtime = javaHome ? await getRuntime(javaHome, { withVersion: true }) : undefined;
if (runtime) {
    javaHome = runtime.homedir;
    javaVersion = runtime.version?.major ?? 0;
} else if (javaRuntimes.length) {
...
```

CsCherrYY: I'm OK if it's cheap. never mind.

PR #873, opened 2022-01-07
use `sass` instead of `node-sass`

### LINKED ISSUE #872: migrate `node-sass` to `dart-sass`
follow https://github.com/microsoft/vscode-java-pack/pull/871#issuecomment-1005441251

CsCherrYY:
fix #872 

The current imports on ttf files will cause some errors.

> SassError: Can't find stylesheet to import

The root cause might be the import statement from `sass` supports only css files, see https://sass-lang.com/documentation/at-rules/import#plain-css-imports. The ttf imports are likely redundant, and removing them brings no change to the existing pages.


Eskibear:
It makes sense because .ttf is already referenced by .css file.  Please double check if codicon icons are correctly rendered (e.g. pencil/edit icon in  Configure Java Runtime Page)  before get it merged.

Eskibear (approved):
Good job!

CsCherrYY:
image

looks like it works as expect.

PR #886, opened 2022-01-18
Support toggling awt development in help center

jdneo:
This PR adds a new entry to toggle the awt development in the Student section of Help Center.

Signed-off-by: sheche <sheche@microsoft.com>


Eskibear (dismissed):
LGTM. 

Just wondering if we can simply add the entry in `private groups`, e.g. make it a getter, and the args value depends on `this.props.isAwtDisabled`. But I'm not sure if that would cause refresh of the whole page after you click the action.

jdneo:
> LGTM.
> 
> Just wondering if we can simply add the entry in `private groups`, e.g. make it a getter, and the args value depends on `this.props.isAwtDisabled`. But I'm not sure if that would cause refresh of the whole page after you click the action.

Good idea! This simplified the code and won't refresh the whole page

Eskibear (approved):

PR #893, opened 2022-01-25
update per new setting java.jdt.ls.java.home

Eskibear:
per upstream change https://github.com/redhat-developer/vscode-java/pull/2284


jdneo (approved):

src/java-runtime/utils/upstreamApi.ts
```diff
@@ -42,51 +42,52 @@ export async function resolveRequirements(): Promise<{
             javaVersion = await getMajorVersion(javaHome);
             toolingJre = javaHome;
```
testforstephen, line 43: On vscode-java, it's a different about picking java.home as tooling jre.

image
See [source code on vscode-java](https://github.com/redhat-developer/vscode-java/blob/de65a586d36a79034a533d564f470b3a2754c1fe/src/requirements.ts#L56-L59)

Eskibear: so in vscode-java, `java.home` will not be picked up as jdtls runtime, right?

Eskibear: BTW, here the logic is to determine whether to pop up "configure Java runtime" for the tooling runtime. Different from upstream vscode-java, here in pack, we simply detect whether toolingJre is valid. Default project runtime `javaHome` is not important.

testforstephen: > so in vscode-java, java.home will not be picked up as jdtls runtime, right?

Yes, here is the order to look for a tooling JRE:
"java.jdt.ls.java.home" > embedded jre > "java.home"

Eskibear: Thanks, updated.

testforstephen (approved):

PR #925, opened 2022-03-03
Add LS daemon to monitor error and performance

Eskibear:


Eskibear:
I'll directly merge it as we've dogfood the private build.

PR #947, opened 2022-03-25
focusing on specific error in server logs

Eskibear:
1. add whitelist of error mssages. Only those on the list will be sent back. Otherwise only tags found in the message will be sent back.
2. After LS service is ready, stop listen to server log changes. So we focus only on startup stage.

Sample Error messages:
- "Error in calling delegate command handler"
- "tags:maven,bundle"
- ""

### DIFF
--- a/src/daemon/serverLog/whitelist.ts
+++ b/src/daemon/serverLog/whitelist.ts
@@ -22,6 +22,7 @@ const MESSAGE_WHITELIST: string[] = [
     "Failed to detect project changes",
     "Failed to update qualified index.",
     "Failed to launch debuggee in terminal. Reason: java.util.concurrent.TimeoutException: timeout",
+    "Failed to save JDT index",
     "failed to send diagnostics",
     "FrameworkEvent ERROR",
     "Initialization failed",
--- a/src/daemon/serverLog/whitelist.ts
+++ b/src/daemon/serverLog/whitelist.ts
@@ -3,6 +3,7 @@
 
 const TAGS: string[] = [
     "buildship",
+    "m2e",
     "gradle",
     "maven",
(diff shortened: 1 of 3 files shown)


src/daemon/serverLog/whitelist.ts
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+const TAGS: string[] = [
+    "buildship",
```
Eskibear, line 5: let me know if you want to add more "keywords"

src/daemon/serverLog/whitelist.ts
```diff
@@ -0,0 +1,51 @@
...
+    "gradle",
+    "maven",
```
jdneo, line 7: what about m2e?

Eskibear: ok, will add

src/daemon/serverLog/whitelist.ts
```diff
@@ -0,0 +1,52 @@
...
+    "BadLocationException",
+    "Could not load Gradle version information",
+    "Error computing hover",
+    "Error in calling delegate command handler",
+    "Error in JDT Core during AST creation",
+    "Error occurred while deleting",
+    "Error filtering index locations based on qualifier.",
+    "Failed to publish diagnostics for",
+    "Failed to detect project changes",
+    "Failed to update qualified index.",
```
testforstephen, line 23: Add one "**Failed to save JDT index**"

Eskibear: added.

PR #974, opened 2022-04-19
Add `Failed to configure project` to whitelist

CsCherrYY:
It's a buildship common error message. see https://github.com/eclipse/buildship/blob/35d01caad3ab023fdd9fdf4a1b9e9c738bff7fd0/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/workspace/BaseConfigurator.java#L68


src/daemon/serverLog/whitelist.ts
```diff
@@ -39,6 +39,7 @@ const MESSAGE_WHITELIST: string[] = [
     "Synchronize Gradle projects with workspace failed due to an error in the referenced Gradle build.",
     "Unable to read JavaModelManager nonChainingJarsCache file",
     "Workspace restored, but some problems occurred.",
+    "Failed to configure project"
```
Eskibear, line 42: alphabetical order, better to move it upward.

Eskibear (approved):

PR #1002, opened 2022-05-25
Add "Open Help Center" step in walkthrough

CsCherrYY:
https://user-images.githubusercontent.com/45906942/170209519-8975a8bb-4786-4950-8eb3-3990f59d0b77.mp4


Eskibear:
Is it designed so to open it in the same colume instead of open-to-side?

silenceJialuo:
@Eskibear ,good question, based on previous talking with UX, it is designed for two different modes,

For day 1 experience, since it is the last step in the walkthrough, users will click the blue button and open a new tab to see the tutorial in the Help center with a full panel. In this way, it may impress them with full panel in same column. If they close the Help center, they can see the walkthrough as well.

For daily experience,  users may open it by Command Palette or shortcuts. And we put it to open-to-side, always in right side because users may concentrate on coding and just use the help center for some tips.

PR #1020, opened 2022-07-06
npm update

Eskibear:

### DIFF
--- a/package.json
+++ b/package.json
@@ -312,12 +312,12 @@
     "@types/fs-extra": "^9.0.13",
     "@types/lodash": "^4.14.182",
     "@types/minimatch": "^3.0.5",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -346,10 +346,10 @@
     "VisualStudioExptTeam.vscodeintellicode"
   ],
   "dependencies": {
-    "@iconify-icons/codicon": "^1.2.8",
+    "@iconify-icons/codicon": "1.2.8",
```
Eskibear, line 349: Every time this lib gets updated, we have to create corresponding license audit request due to governance. Here I simply pin this version because we don't use new icons from newer versions. We can update it later on demand.

jdneo (approved):

PR #1025, opened 2022-07-13
Simplify the clean cache notification buttons

testforstephen:
Since the "Later" button doesn't provide much help, remove it.

Before:
image

After:
image


Eskibear:
how do I reproduce and verify this feature? @testforstephen

testforstephen:
I kept the scene of my macOS when I met this problem once. It's not easy to manually reproduce it. This's just a minor change, it's safe to skip the verification.

Eskibear:
OK, then I'll proceed to release.

PR #1063, opened 2022-09-27
stop logging useless info when embedded jre is absent

Eskibear:


Eskibear:
When running redhat.java universal version without an embedded jre, it repeatedly print error messages into console, which spams the console and disturb debugging of other components. 

Here to mute it as it doesn't provide much useful info.

jdneo (approved):

PR #1101, opened 2022-11-03
Remove Release Notes from Overview

Sti2nd:
Hasn't been updated since 2019, so I thought it would simply be best to remove the Release Notes section.


Eskibear (approved):
Thank you.  It makes sense to remove them, as we now publish release notes here: 
https://devblogs.microsoft.com/search?query=Java+on+Visual+Studio+Code

PR #1147, opened 2023-02-07
Avoid the LSP request name get redacted

jdneo:


Eskibear (dismissed):

Eskibear:
why not just only process logs in latest session just like before? 
See https://github.com/microsoft/vscode-java-pack/pull/1145/files#diff-c618578402e9fdd35d1287b5f6d84018fb7b57267689a90d0f28a77091a79898L58-L59

Eskibear (approved):

PR #1154, opened 2023-02-13
Aggregate the perf traces

jdneo:
- to avoid sending to many traces during startup.


Eskibear (approved):
LGTM
BTW, `number` type will go to "measures" automatically instead of properties.

PR #1161, opened 2023-03-05
pivotal.vscode-boot-dev-pack -> vmware.vscode-boot-dev-pack

deftdawg:
`pivotal.vscode-boot-dev-pack` is no longer available in the extensions market, it has been replaced by `vmware.vscode-boot-dev-pack`


deftdawg:
@microsoft-github-policy-service agree

Eskibear (approved):
Good catch! Thanks!

PR #1168, opened 2023-03-16
monitor session uptime when cleaning JavaWS

Eskibear:


src/extension.ts
```diff
@@ -99,9 +101,12 @@ function initializeTelemetry(_context: vscode.ExtensionContext) {
 
 export async function deactivate() {
   if (cleanJavaWorkspaceIndicator && fs.existsSync(cleanJavaWorkspaceIndicator)) {
+    const now = Date.now();
+
     sendInfo("", {
       name: "cleanJavaLSWorkspace",
-      timestamp: Date.now().toString()
+      timestamp: now.toString(),
+      time: now - activatedTimestamp
```
jdneo, line 109: Use `performace.now()`? For a long running session, it will have more chance to be affected if user adjusts the systom clock.

Eskibear: Sure, we should update it. I suggest that you update all occurence of `Date.now()` in a separate PR. For the moment, I don't want to update it.

jdneo (approved):

PR #1177, opened 2023-03-21
Exclude invalid completion requests during tracing

testforstephen:


jdneo (approved):
There can be two choices here:

1. eliminate the event whose result length is undefined.
2. send a telemetry for those undefined-length event, in case in the future we need the data.

Both are fine to me, so I would leave the choice to you.

Issue #1169, opened 2023-03-16
update all occurence of `Date.now()` to `performance.now` 


Eskibear:
Sure, we should update it. I suggest that you update all occurence of `Date.now()` in a separate PR. For the moment, I don't want to update it.

_Originally posted by @Eskibear in https://github.com/microsoft/vscode-java-pack/pull/1168#discussion_r1138211501_

Eskibear:
BTW, an integer of millis is enough I think, remember to perform conversion.

PR #1196, opened 2023-05-05
Use performance.now() to replace Date.now()

jdneo:
related with https://github.com/microsoft/vscode-java-pack/issues/1169

_I found some abnormal data (negative or extremely large)._

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -43,7 +43,7 @@ async function initializeExtension(_operationId: string, context: vscode.Extensi
   initRecommendations(context);
   initDaemon(context);
 
-  activatedTimestamp = Date.now();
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -100,11 +100,10 @@ function initializeTelemetry(_context: vscode.ExtensionContext) {
 }
 
 export async function deactivate() {
-  const now = Date.now();
+  const now = performance.now();
   const data = {
     name: "sessionStatus",
-    timestamp: now.toString(),
```
jdneo, line 106: redundant since each telemetry already have timestamp field.

testforstephen (approved):

PR #1202, opened 2023-05-29
Track e2e startup time

jdneo:


src/extension.ts
```diff
@@ -105,11 +106,36 @@ export async function deactivate() {
...
+  // if it's already activated, we are not clear if it is activated with pack at
+  // the same moment.
+  if (javaExt && !javaExt.isActive) {
+    const delay = promisify(setTimeout);
+    const timeout = 1 * 60 * 1000; // wait 1 min at most
+    let count = 0;
+    while(!javaExt.isActive && count < timeout) {
+      await delay(1000);
+      count += 1000;
+    }
```
testforstephen, line 130: The condition branch `if (javaExt && !javaExt.isActive)` should end at L130.

jdneo: I was intended to do this, while not sure if this makes enough sense. What do you think about the follow case:

1. In an empty workspace, user triggers some `redhat.java` specific command. such as
image to open a webview.
2. Then, `redhat.java` will be activated, while pack is not.

So the purpose here is to make sure that two extensions are almost activated at the same time.

MrG9090 (approved):

PR #1243, opened 2023-09-21
Record LSP request aggregation on extension exit

testforstephen:


jdneo (dismissed):

src/daemon/index.ts
```diff
@@ -103,21 +96,45 @@ async function traceLSPPerformance(javaExt: vscode.Extension<any>) {
...
    }
 
-   lspUsageStats = new LSPUsageStats(javaExtVersion, sampling);
+   // Enable it since redhat.java@1.23.0
+   if (javaExt.exports?.onWillRequestStart) {
+      lspUsageStats = new LSPUsageStats(javaExtVersion, sampling);
+      try {
+         // Load HdrHistogramJS WASM module
+         if (vscode.env.uiKind === vscode.UIKind.Desktop) {
+            hdr.initWebAssemblySync();
```
akaroml, line 105: Any specific reason to use the sync version?

src/daemon/index.ts
```diff
@@ -174,7 +191,7 @@ async function traceLSPPerformance(javaExt: vscode.Extension<any>) {
       }
 
       if (traceEvent.resultLength === 0) {
```
akaroml, line 193: How about the isEmpty check to also cover empty objects?

akaroml (commented):
2 questions.

src/daemon/index.ts
testforstephen: This is the recommended approach from the library. On server side, we can use initAsync directly, but for browser, use the async.

https://github.com/HdrHistogram/HdrHistogramJS#boosting-performances-with-webassembly

src/daemon/index.ts
testforstephen: resultLength is not the raw response object. It's a processed value parsed from the response.

See
https://github.com/redhat-developer/vscode-java/blob/812613b5923ba27508f0a15a605f4cf80faa2388/src/TracingLanguageClient.ts#L134

akaroml (approved):
LGTM!

PR #1248, opened 2023-10-07
Differentiate the LSP stats between syntax server and standard server

testforstephen:


src/daemon/index.ts
```diff
@@ -319,6 +324,86 @@ export function sendLSPUsageStats() {
...
+   public constructor(readonly javaExtVersion: string, readonly sampling: string) {
+      this.lspStats = new LSPUsageStats(javaExtVersion, sampling, false);
+      // These pre-release versions include the fromSyntaxServer property in the requestEnd event,
+      // but not in the requestStart event.
+      const ignoreVersions = [
+         "1.24.2023100604",
+         "1.24.2023100504",
+         "1.24.2023100404",
+      ];
+      if (sampling !== "pre-release" || !ignoreVersions.includes(javaExtVersion)) {
```
jdneo, line 340: Add a `TODO` marker here to update the logic before stable release? (if I understand correctly, the code here is to temporarily collect syntax server traces via pre-release channel)

jdneo (approved):

src/daemon/index.ts
testforstephen: done.

jdneo (approved):

PR #1263, opened 2023-11-24
Support update project jdk

jdneo:


testforstephen:
Overall works well. One corner case is that when I edit the setting 'java.configuration.runtimes', the JDK list in Classpath Configuration page is not refreshed. I have to close it and reopen it to get the latest JDK list. The suggestion is that you can listen at the configuration change event of 'java.configuration.runtimes', re-sync the jdk list.

testforstephen (approved):

PR #1315, opened 2024-04-26
add `inspectRange`/`inspectClass` based on `inspectCode`.

wangmingliang-ms:
* inspecting class/symbol will both go to `inspectRange`
* for random selections, we will adjust(expand) it to the union range of container symbols first before inspecting it. e.g. if user selection covers bodies of two methods, then the range will be expanded to the full range of these two methods.

### DIFF
--- a/src/copilot/inspect/InspectionCopilot.ts
+++ b/src/copilot/inspect/InspectionCopilot.ts
@@ -1,7 +1,9 @@
 import { instrumentSimpleOperation, sendInfo } from "vscode-extension-telemetry-wrapper";
 import Copilot from "../Copilot";
-import { logger } from "../utils";
+import { getContainedClassesOfRange, getContainingClassOfRange, getIntersectionMethodsOfRange, getUnionRange, logger } from "../utils";
 import { Inspection } from "./Inspection";
+import path from "path";
+import { TextDocument, DocumentSymbol, SymbolKind, ProgressLocation, Position, Range, Selection, window } from "vscode";
 
 export default class InspectionCopilot extends Copilot {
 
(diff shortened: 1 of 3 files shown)


src/copilot/utils.ts
```diff
@@ -1,3 +1,52 @@
-import { LogOutputChannel, window } from "vscode";
+import { DocumentSymbol, LogOutputChannel, SymbolKind, TextDocument, commands, window, Range, Selection } from "vscode";
+
+export const CLASS_KINDS: SymbolKind[] = [SymbolKind.Class, SymbolKind.Interface, SymbolKind.Enum];
+export const METHOD_KINDS: SymbolKind[] = [SymbolKind.Method, SymbolKind.Constructor];
 
 export const logger: LogOutputChannel = window.createOutputChannel("Java Rewriting Suggestions", { log: true });
+
+export async function getContainedClassesOfRange(range: Range | Selection, document: TextDocument): Promise<DocumentSymbol[]> {
```
testforstephen, line 8: It's not easy to understand the name `getContainedClassesOfRange` and `getContainingClassOfRange`.

-> getClassesInRange ?

src/copilot/utils.ts
```diff
@@ -1,3 +1,52 @@
...
+export async function getContainingClassOfRange(range: Range | Selection, document: TextDocument): Promise<DocumentSymbol> {
+    const symbols = await getClassesAndMethodsOfDocument(document);
+    return symbols.filter(symbol => CLASS_KINDS.includes(symbol.kind))
+        // reverse the classes to get the innermost class first
+        .reverse().filter(clazz => clazz.range.contains(range))[0];
+}
+
+export function getUnionRange(symbols: DocumentSymbol[]): Range {
+    let result: Range = new Range(symbols[0].range.start, symbols[0].range.end);
+    for (const method of symbols) {
```
testforstephen, line 29: method -> symbol ? Calling the name as method is a little weird.

src/copilot/inspect/InspectionCopilot.ts
```diff
@@ -139,6 +196,19 @@ export default class InspectionCopilot extends Copilot {
         });
     }
 
+    private async doInspectRange(document: TextDocument, range: Range | Selection): Promise<Inspection[]> {
+        const adjustedRange = new Range(new Position(range.start.line, 0), new Position(range.end.line, document.lineAt(range.end.line).text.length));
+        const content: string = document.getText(adjustedRange);
+        const offset = range.start.line;
```
testforstephen, line 202: offset -> startLine ?

offset is usually the absolute position in the document text, and offset can be converted to line & column position.

testforstephen (changes requested):

src/copilot/utils.ts
wangmingliang-ms: I'll add javadoc to explain these methods.

testforstephen: it's better to use the method name to explain itself rather than the doc.

src/copilot/inspect/InspectionCopilot.ts
wangmingliang-ms: renamed

src/copilot/utils.ts
wangmingliang-ms: `getClassesInRange` can not explain what it does exactly (I added js doc), and lose symmetry with  `getContainerClassOfRange`. I would like to keep `getContainedClassesOfRange` and `getContainerClassOfRange`.

wangmingliang-ms: renamed as `getClassesContainedInRange` and `getInnermostClassContainsRange` to get some symmetry.

testforstephen (approved):

PR #1320, opened 2024-05-07
cache inspections by symbols

wangmingliang-ms:
so that inspections of other symbols are still valid even if one symbol is modified.


testforstephen:
No idea on how the cache will be used. It's better to include a full small feature in a PR.

src/copilot/inspect/index.ts
```diff
@@ -1,21 +1,20 @@
...
+    registerCommands(renderer);
 
     context.subscriptions.push(
-        workspace.onDidOpenTextDocument(doc => inspectActionCodeLenses.rerender(doc)), // Rerender class codelens when open a new document
-        workspace.onDidChangeTextDocument(e => inspectActionCodeLenses.rerender(e.document)), // Rerender class codelens when change a document
-        languages.registerCodeActionsProvider({ language: 'java' }, { provideCodeActions: rewrite }), // add code action to rewrite code
+        languages.registerCodeActionsProvider({ language: 'java' }, { provideCodeActions: fixDiagnostic }), // Fix using Copilot
+        languages.registerCodeActionsProvider({ language: 'java' }, { provideCodeActions: rewrite }), // Inspect using Copilot
+        workspace.onDidOpenTextDocument(doc => renderer.rerender(doc)), // Rerender class codelens and cached suggestions on document open
+        workspace.onDidChangeTextDocument(e => renderer.rerender(e.document, true)), // Rerender class codelens and cached suggestions debouncely on document change
```
testforstephen, line 14: When the text document is changed, shouldn't you re-inspect it again before re-render?

testforstephen (changes requested):

src/copilot/inspect/index.ts
wangmingliang-ms: `inspect` is expensive. Classes usually have multiple methods, if user modifies one method, inspections of other methods are still available (that is why we have cache and why we don't need to re-inspect the whole document again), but their positions may be changed (that's why we need to rerender).

src/copilot/inspect/InspectionCache.ts
```diff
@@ -0,0 +1,94 @@
...
+import { METHOD_KINDS, getClassesAndMethodsOfDocument, logger } from '../utils';
+import { Inspection } from './Inspection';
+import * as crypto from "crypto";
+import { SymbolNode } from './SymbolNode';
+
+/**
+ * A map based cache for inspections of a document.
+ * format: `Map<documentKey, Map<symbolQualifiedName, [symbolVersionId, Promise<Inspection[]>]`
+ */
+const DOC_SYMBOL_VERSION_INSPECTIONS: Map<string, Map<string, [string, Promise<Inspection[]>]>> = new Map();
```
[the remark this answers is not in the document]
wangmingliang-ms: You are right.
in my old implementation, the `inspecting` process is saved, so that next call will get the promise directly when first request is still processing (instead of sending another request). but it's no need now

src/copilot/inspect/InspectionCache.ts
```diff
@@ -0,0 +1,90 @@
...
+                s.problem.position.line = s.problem.position.relativeLine + symbol.range.start.line;
+            });
+            return inspections;
+        }
+        logger.debug(`cache miss for ${SymbolKind[symbol.kind]} ${symbol.qualifiedName} of ${document.uri.fsPath}`);
+        return [];
+    }
+
+    public static cache(document: TextDocument, symbols: SymbolNode[], inspections: Inspection[]): void {
+        for (const symbol of symbols) {
```
testforstephen, line 57: this updates the cache with the latest symbols in the document, but never has the chance to remove the outdated symbols from the cache. This is not like a sensible cache design to me.

src/copilot/inspect/index.ts
testforstephen: the question is if I introduce a new method to the file, when will it update the inspections?

src/copilot/inspect/InspectionCache.ts
wangmingliang-ms: changed to **clear cache when user close a document**.
but I don't think it's necessary to clear inspection cache for outdated symbols because it's common case that user make some changes to the code and then ctrl-z to undo them (especially when demo), I keep these inspection cache to avoid repeat inspecting.

testforstephen (approved):

PR #1333, opened 2024-05-21
error handling, messaging and refactoring.

wangmingliang-ms:
1. only activate the copilot related feature on 1.90.0-insider or later.
2. simplify messages/logs.
3. show error message on command failures.
4. retry when failed to getting java settings (if inspect before project loaded)
5. checking extension installed using event instead of polling.

### DIFF
--- a/src/copilot/Copilot.ts
+++ b/src/copilot/Copilot.ts
@@ -28,7 +28,8 @@ export default class Copilot {
         const messages = [...systemMessagesOrSamples];
         const _send = async (message: string): Promise<boolean> => {
             rounds++;
-            logger.info(`User: \n`, message);
+            logger.debug(`User: \n`, message);
+            logger.info(`User: ${message.split('\n')[0]}...`);
             messages.push(new LanguageModelChatMessage(LanguageModelChatMessageRole.User, message));
             logger.info('Copilot: thinking...');
 
@@ -47,7 +48,8 @@ export default class Copilot {
                 throw new Error(`Failed to send request to copilot: ${e}`);
             }
             messages.push(new LanguageModelChatMessage(LanguageModelChatMessageRole.Assistant, rawAnswer));
-            logger.info(`Copilot: \n`, rawAnswer);
+            logger.debug(`Copilot: \n`, rawAnswer);
+            logger.debug(`Copilot: ${rawAnswer.split('\n')[0]}...`);
             answer += rawAnswer;
             return answer.trim().endsWith(this.endMark);
         };
--- a/src/copilot/Copilot.ts
+++ b/src/copilot/Copilot.ts
@@ -49,7 +49,7 @@ export default class Copilot {
             }
             messages.push(new LanguageModelChatMessage(LanguageModelChatMessageRole.Assistant, rawAnswer));
             logger.debug(`Copilot: \n`, rawAnswer);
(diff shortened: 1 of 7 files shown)


src/copilot/inspect/InspectionCopilot.ts
```diff
@@ -220,10 +220,8 @@ export default class InspectionCopilot extends Copilot {
         const adjustedRange = new Range(new Position(range.start.line, 0), new Position(range.end.line, document.lineAt(range.end.line).text.length));
         const content: string = document.getText(adjustedRange);
         const startLine = range.start.line;
-        const javaVersion = await getProjectJavaVersion(document);
-        const inspections = await this.inspectCode(content, {
-            javaVersion
-        });
+        const context = await this.collectProjectContext(document);
+        const inspections = await this.inspectCode(content, context);
```
testforstephen, line 224: First glance, we just thought `content` and `context` are the same word. Maybe rename `context` -> `javaContext` ?

src/copilot/utils.ts
```diff
@@ -85,6 +86,32 @@ export function isCodeLensDisabled(): boolean {
...
+                throw error;
+            } else {
+                await new Promise(resolve => setTimeout(resolve, retryInterval));
+            }
+        }
+    }
+}
+
+export function isLlmApiReady(): boolean {
+    return new SemVer(version).compare(new SemVer("1.90.0-insider")) >= 0;
```
testforstephen, line 116: do you think if it is necessary to compare "1.90.0" as well?

src/copilot/Copilot.ts
```diff
@@ -47,7 +48,8 @@ export default class Copilot {
                 throw new Error(`Failed to send request to copilot: ${e}`);
             }
             messages.push(new LanguageModelChatMessage(LanguageModelChatMessageRole.Assistant, rawAnswer));
-            logger.info(`Copilot: \n`, rawAnswer);
+            logger.debug(`Copilot: \n`, rawAnswer);
+            logger.debug(`Copilot: ${rawAnswer.split('\n')[0]}...`);
```
testforstephen, line 52: should be `logger.info` ?

src/copilot/utils.ts
wangmingliang-ms: as far as I know, `1.90.0` is greater/newer than `1.90.0-insider`

src/copilot/inspect/InspectionCopilot.ts
wangmingliang-ms: I will rename it to `projectContext` because it may contains information not related to java.

src/copilot/utils.ts
wangmingliang-ms: confirmed!

testforstephen (approved):

src/copilot/inspect/DocumentRenderer.ts
```diff
@@ -104,7 +104,11 @@ export class DocumentRenderer {
         if (settings.length === 0) {
             settings.push('diagnostics');
             settings.push('rulerhighlights');
-            settings.push(isCodeLensDisabled() ? 'guttericons' : 'codelenses');
+            const disabled = isCodeLensDisabled();
+            if (disabled) {
+                logger.warn('CodeLens is disabled, fallback to GutterIcons');
```
akaroml, line 109: I think we can fallback to Comments UI.

PR #1335, opened 2024-05-22
Add compiler section in Project Setting page

jdneo:
requires: https://github.com/eclipse-jdtls/eclipse.jdt.ls/pull/3162


testforstephen:
image

It can be confusing when the bytecode list displays only a partial list of versions. To enhance usability, consider supporting all Java versions. Additionally, if a user sets a version higher than the project’s JDK runtime, prompt them to upgrade the project JDK as well.

testforstephen (approved):

PR #1372, opened 2024-07-26
Add code action section into walkthrough

jdneo:
image

The `or` is not added finally, because it looks wired.


package.nls.json
```diff
@@ -7,6 +7,8 @@
 
     "walkthrough.java.projects.inactive": "Java Project Explorer helps you to manage your projects and dependencies. Create a new project or open an existing folder to see the full features.\n[Open Command Palette](command:workbench.action.quickOpen?%22>java%22)\n__Try searching for 'Create Java Project' or 'Open Folder'__",
 
+    "walkthrough.java.codeActions": "Find available actions for your code in the editor:\n  - Open the right-click menu then select \"Refactor...\" or \"Source Action...\".\n  - Click on the light bulb💡.",
+
```
testforstephen, line 11: Add `OR` between two options. This `OR` can occupy a separate line.

nickzhums (approved):

package.nls.json
nickzhums: Sheng and I tested this approach and found it looked really strange. We talked with UX and think the original design actually looks better

jdneo: image

testforstephen (approved):

PR #1383, opened 2024-08-20
fix: Extension failed to activate

jdneo:
- This PR fixes the issue that the pack extension will fail to activate when the copilot required extensions are not installed or activated.


wangmingliang-ms (approved):
LGTM

PR #1387, opened 2024-09-03
Implement LM tools to provide project context in chat

jdneo:


src/copilot/context/JavaLanguageModelTool.ts
```diff
@@ -0,0 +1,38 @@
...
+// Licensed under the MIT license.
+
+import { JavaProject, JavaProjectContext } from "./JavaProject";
+import * as vscode from "vscode";
+import { logger } from "../logger";
+import { fixedInstrumentSimpleOperation, sendEvent } from "../utils";
+import { JavaProjectContextResult } from "./results/JavaProjectContextResult";
+import { EmptyContextResult } from "./results/EmptyContextResult";
+
+export class JavaContextProvider implements vscode.LanguageModelTool {
```
wangmingliang-ms, line 11: name of the file and the class should better be consistent.

wangmingliang-ms (changes requested):
suggest renaming the file.

src/copilot/context/JavaLanguageModelTool.ts
```diff
@@ -0,0 +1,38 @@
...
+    invoke(_options: vscode.LanguageModelToolInvocationOptions, _token: vscode.CancellationToken): Thenable<vscode.LanguageModelToolResult> {
+        return fixedInstrumentSimpleOperation("java.copilot.vscodeJavaContextTool.invoke", async () => {
+            logger.info(`Invoking LM Tool "vscodeJavaContextTool"...`);
+            const document = vscode.window.activeTextEditor?.document ?? vscode.workspace.textDocuments[0];
```
wangmingliang-ms, line 15: we should use selected file to build the project instance in case `#file:xxx` variable is used, but I doubt if copilot has such API to get the selected file.

wangmingliang-ms (approved):
LGTM

PR #1400, opened 2025-01-10
English grammar

hippietrail:
Fix verb agreement with plural noun: There is no .. projects→There are no ... projects

Alternatively maybe it should be: `There is no Java project opened in the current workspace`


testforstephen (approved):
LGTM. Thanks for contribution.

